PR #270, opened 2014-06-09
addHTML fixes for PHP 5.3

hskrtich:
I found 2 bugs with the addHTML function. One when running under PHP 5.3 the default character encoding is ISO-8859-1 (This was changed to UTF-8 in PHP 5.4) which causes loadXML to hork because of it converts things like & nbsp ; to non-valid UTF-8 characters. The second is when you try to use an ampersand in the html you are trying to convert to docx.

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -42,9 +42,14 @@ class Html
          * which could be applied when such an element occurs in the parseNode function.
          */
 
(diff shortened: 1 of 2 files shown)


AITG:
Is this issue of embedded & definitely fixed as we're having issues with it, I've had to convert them to ' and ' in order to get the doc generated?

We're Using 0.12.0 of phpword and PHP 5.3.10.

PR #304, opened 2014-07-03
Image from html to docx

basjan:
Added images from html, including inline base64 image data.
Some styling, i.e. width, height, float:left and float:right


troosan:
this has been implemented in various other PR (#937, #934)

PR #330, opened 2014-08-09
READ_EQUAL must return FALSE if expected not found

iliepandia:
In the previous version of the code self:READ_EQUAL would return the value found in the style if the expected value was not found. That this meant was that invalid NON-boolean values would be set for superScript or subScript when a match would not be found.


troosan:
already done

PR #346, opened 2014-08-24
Adds an ability to set custom variable delimiters inside Template class

Strate:
1. Added reference to parent from Element class.
   0.5: Added column number in Cell class.
2. Added ability to set custom variable delimiters
3. Added ability to use spaces between variable delimiter and variable name

Sorry for russian comments in commits :)


troosan:
too old to merge.
also, too many different changes in single PR, feel free to create new PR

PR #362, opened 2014-09-08
Centimeter to twip convertion was inaccurate

tostercx:
1 inch = 2.54 cm = 1440 twips
so
1 cm = 1440/2.54 = 566.93 twips

Not sure where the previous value (565.217) came from. I noticed margins being off in word by 2-5 tenths of a mm. Not that it has much practical concerns but hovering over a margin that's supposed to be exactly 1 cm and finding out it's 0.97 cm wasn't nice.


troosan:
Not applicable anymore. Class does not exist anymore.

PR #369, opened 2014-09-16
Add whitespace between Tag CleanUp

philippfrenzel:
Added a white clean between Tags -> as lists are normally rendered with a double newline inbetween


troosan:
I guess a better way would be to set
```php
$dom->preserveWhiteSpace = false;
```
I've added this in the current development branch (with an additional optional param to keep backward compatibility)

philippfrenzel:
Hi @troosan ,

thanks for adding the hint with preserveWhiteSpace -> what about the addition of the <i>-Tag?

Thanks Philipp

troosan:
@philippfrenzel that has already been done in de development branch

philippfrenzel:
ok - thanks for the update - you are doing a great job!

PR #371, opened 2014-09-16
Update bootstrap.php

franzholz:
The phpunit error message should give the full filename as the error details.


RomanSyroeshko:
Thanks for the idea.

PR #380, opened 2014-09-23
Remove newlines from the HTML table cells 

franzholz:
I use the example which creates the file Sample_07_TemplateCloneRow.docx. However in HTML and PDF you see the newlines in the table cells.

``` html
<tr>
<td>
<p></p>
<p><span style="font-family: 'Calibri';">Value 1</span></p>
<p><span style="font-family: 'Calibri';">:</span></p>
</td>
<td>
<p><span style="font-family: 'Calibri'; font-weight: bold;">Sun</span></p>
</td>
</tr>
```

This patch will remove these `<p>` s inside of the `<td>` tags. Then the table cells in the docx will look the same as in the generated HTML and PDF files.


troosan:
This doesn't seem to be applicable anymore.
I'm running Sample_09_Tables.php and it looks fine.

Issue #383, opened 2014-09-25
Strange behavior of cloneRow with multiple tables


scmember:
Found a strange behavior of cloneRow. I use a sample from phpWord to demonstrate it. I just copy a second table, delete replaceable tags and fill the space between tables with some text. Clone function is not work properly, it clones all the content after original table. Why?

Here's the link to modified sample: https://dl.dropboxusercontent.com/u/102563812/Sample_07_TemplateCloneRow.docx
Attached image shows how cloneRow works with this modified sample
![image 6](https://cloud.githubusercontent.com/assets/5699583/4400253/a7cfc7dc-447d-11e4-9ca8-ad917c862fc5.jpg)

Thanks!

PR #392, opened 2014-10-04
Bugfix for #383: Strange behavior of cloneRow with multiple tables

scmember:
Added a check for the end of table, for correct processing of multiple tables with spanned rows

### DIFF
--- a/src/PhpWord/Template.php
+++ b/src/PhpWord/Template.php
@@ -203,10 +203,11 @@ class Template
                     break;
                 }
 
-                // If tmpXmlRow doesn't contain continue, this row is no longer part of the spanned row.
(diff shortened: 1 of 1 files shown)


FBnil:
@scmember Can you close this pull request, problem has been identified and I have a patch request ready.

Your proposal for a fix is not quite correct, it would match MSWord docx documents, but not LibreOffice docx documents due to them not having extra spaces after a close (`/>`). 
i.e. This works:  `preg_match('#<w:vMerge ?/>#', $tmpXmlRow)` (idem with the others preg_matches)

Because of this, tables made with OpenOffice would never match and go outside the table. Making cloneRow() not that useful when w:vMerge was encountered. 

Superseded by https://github.com/PHPOffice/PHPWord/pull/1147

scmember:
@FBnil agree, close this pr

PR #410, opened 2014-11-06
Add is safe parameter to setValue method

franciscocardoso:
This commits makes it possivel to set valid docx xml into a template
variable.


FBnil:
@franciscocardoso good code and very necessary in 2014, but your functionality has been superseded by this code:
```
if (Settings::isOutputEscapingEnabled()) {
            $xmlEscaper = new Xml();
            $replace = $xmlEscaper->escape($replace);
        }
```
with a global parameter that can be enabled/disabled. Can you close your pull request?

PR #419, opened 2014-11-12
Implement styles for tables when parsing Html

ugomeda:
This pull request ehances the `\PhpOffice\PhpWord\Shared\Html::addHtml` function to :
- Be able to use tables
- Change the default styles
- Use colspan (sorry, I didn't see any reliable way to implement rowspan fonctionnality at the moment, mainly due to the lack of end of node event in the implementation)


danilocarva9:
Can't use rowspan on addHtml as well, it's august 2019... still no fix :(

PR #432, opened 2014-11-20
Update Template.php

alexisQuiroz:
method for analyzing an array of variables within a block and retain the block to insert another arrangement of variables. insert different variables in different blocks, multiple variables in multiple blocks.

### DIFF
--- a/src/PhpWord/Template.php
+++ b/src/PhpWord/Template.php
@@ -223,7 +223,34 @@ class Template
 
         $this->documentXML = $result;
     }
+    
+ /*
+     * write multi variable in mutiBlock
+     * @param string $blockname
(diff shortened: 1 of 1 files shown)


FBnil:
I see merit in a scoped Block (which unfortunately, it does not), but it has been superseded by 0.13.0, doing a `cloneBlock("blockname", 1); setValue(array_keys($a), array_values($a));`
The first command gets rid of the blocks, but lets the content of the block stay, and the second does a multi value named array insertion, where `$a = ['foo'=>'bar', 'baz'=>42];`

In 2014 this was valuable code. But with matching functionality in 0.13.0 it can be closed.

@alexisQuiroz : Aún por aquí?

PR #459, opened 2015-01-01
Extend `cloneBlock` function to use like `cloneRow` but without table

totpero:
I  have extend `cloneBlock` function, i have add another parameter `$incrementVariables` default TRUE;

This parameter can transform the `cloneBlock` function in alternative to `cloneRow` function working without table;

I extend this because i have need to use the `cloneRow` but in regular text not using table and `cloneRow` not allow this.


FBnil:
I did an effort too to add this functionality, see https://github.com/FBnil/PHPWord/pull/1
I incorporated some of your ideas and expanded on that.

troosan:
this one will be hard to merge because a lot changed since.
I'll merge #817 instead, but it needs a bit of work to be able to pass argument $incrementVariables (which should default to false for backward compatibility reasons)

PR #467, opened 2015-01-08
Fix bug in getVariables

supertanuki:
Clean the variables name before return them (they may contain xml tags)

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -161,6 +161,9 @@ class TemplateProcessor
         foreach ($this->temporaryDocumentFooters as $footerXML) {
             $variables = array_merge($variables, $this->getVariablesForPart($footerXML));
         }
+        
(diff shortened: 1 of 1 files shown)


FBnil:
@supertanuki Thanks for your time and code.

Instead of patching the code visually, I made `getVariablesForPart` more robust by NOT allowing `<` or `>` , i.e. `preg_match_all('/\$\{([^<>{}]*?)}/i', $documentPartXML, $matches);`  Is that not a better way to fix the problem? Also, did `fixBrokenMacros` not exist back when you edited the code? Should that not have fixed the problem? (or did you disable it, because it was too buggy?)
I would be very interested in a **malformed document** where your fix was necessary to hold it against my new re-implementation of TemplateProcessor. Do you have such a document? And could you still open the document after edit it? (to add that as a testcase)

troosan:
closing as this should not be problem anymore as the fixBrokenMacros should not let such variables pass anymore

Issue #468, opened 2015-01-08
PhpOffice\PhpWord\Element\AbstractContainer error suppression


daum:
In AbstractContainer.php on line 101 right now there is the following:
```php
@list($count, $fontStyle, $paragraphStyle) = $args; // Suppress error
```

The issue is if a user has listeners for warnings/errors that they will still get the error which then can cause their scripts to send extra alert emails.  Switching this to the following would have the same affect, just never produce the warnings to the listeners:

```php
if (isset($args[0])) {
    $count = $args[0];
}
else {
    $count = null;
}
// Same for rest of variables
```

I'll submit a PR with the full fix.  While this does have a few more lines of code, it will allow error monitoring to continue without false positives.

GMTA:
> The issue is if a user has listeners for warnings/errors that they will still get the error which then can cause their scripts to send extra alert emails.

Yeah, if you use [set_error_handler()](http://php.net/set_error_handler) then you should also read the remark about it bypassing [error_reporting()](http://php.net/manual/en/function.error-reporting.php) completely. This is not PHPWord at fault with error suppression, this is your custom error handling ignoring the error level.

Fix your code:
```PHP
if (!(error_reporting() & $errno)) {
    return;
}
```

PR #469, opened 2015-01-08
Issue #468 Error suppression in AbstractContainer update for error listeners

daum:
This pull request addresses #468 so that people who have custom handlers for php errors and warnings will no longer get triggered when using the addTextBreak on sections in certain scenarios where they don't pass all the arguments.  The error suppression stops the error from being shown but it is still passed to the handlers creating unwanted alerts.

I've moved it to a couple of conditionals so that no warnings are created.


troosan:
already fixed in commit 4084d4f191202c40789c4b0396122e353a019106

PR #487, opened 2015-03-06
Added support for word "mail merge" fields

cjbaar:
Recently used this library for a project which required inline text substitution. Legacy document was already built using pre-defined merge fields, and given it's large size, we did not want to reformat to take advantage of the existing setValue() method. I created some new functions to allow replacement of text using the pre-defined merge fields, allowing the same document to be converted using both the older manual method and the new programmatic (PHPWord) method. This is based on a limited set of Word merge documents, but so far all replacements are successful.

### DIFF
--- a/docs/templates-processing.rst
+++ b/docs/templates-processing.rst
@@ -23,3 +23,27 @@ multirow from a single row in a template by using ``TemplateProcessor::cloneRow`
 See ``Sample_23_TemplateBlock.php`` for example on how to clone a block
 of text using ``TemplateProcessor::cloneBlock`` and delete a block of text using
 ``TemplateProcessor::deleteBlock``.
+
+
+
+Merge field processing
+======================
+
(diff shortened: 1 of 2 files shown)


gfobe:
The RegEx searching for `w:instr=" MERGEFIELD` does not trigger in my document [Placeholder.docx](https://github.com/PHPOffice/PHPWord/files/1016386/Placeholder.docx)

cjbaar:
@gfobe: Your document is using a merge field format that we did not encounter on our project. It looks like there are two different types of merge field in Word: simple and complex. We currently only have used the complex method. According to what I have read, if there is only one merge field on a document, it default to simple. As a short work-around, try adding a second merge field in the document and see if that changes things. (You should see fldChar tags instead of fldSimple)

FBnil:
See also https://github.com/PHPOffice/PHPWord/pull/1050

PR #493, opened 2015-03-13
Add Img parser for Html reader

fr05t1k:
HTML reader can parse img tags


FBnil:
Not sure why it hasn't been added (does it work with `./src/PhpWord/Writer/HTML/Element/Image.php`  no extra modifications needed?)

fr05t1k:
@FBnil Try this. I don't remember it.

troosan:
similar PR #934, which is parsing more attributes

PR #498, opened 2015-03-15
change TemplateProcessor to protected

angelomaragna:
I needed to extend class in my own project to merge docx templates.
I found private classes and methods o huge obstacle.
Using protected will give more flexibility to contributors.


RomanSyroeshko:
Good suggestion. 

As soon as you PR points to ``master`` branch instead of ``develop``, I had to implement the changes by myself.

PR #511, opened 2015-03-25
Modified for adding compatible objects

adiabvs:
Issue - OLEobject creation not done properly and hopefully we cannot do. so we can embed object as package.

Because of xls, doc, and ppt types are not packages they are Removed.

Now we can add object of type xlsx,docx and pptx through addObject.


troosan:
indeed, you are right.
But with this fix you broke the behaviour for the "old" xls, doc and ppt.
I'll try to see if we cannot support both format, and maybe also pdf as requested in #1141

PR #516, opened 2015-04-01
Adding condition for document template saving using libreoffice

AntonioMalatesta:
When  I save document template with libreoffice (in .docx format) the cloneRow with nested table doesn't work.
The problem is that the regular expression, that used for search if row is no longer part of the spanned row, was wrong: 
<w:vMerge w:val="continue" /> became <w:vMerge w:val="continue"/> without space at the closure tag.
Sorry for my English.

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -204,7 +204,8 @@ class TemplateProcessor
                 // If tmpXmlRow doesn't contain continue, this row is no longer part of the spanned row.
                 $tmpXmlRow = $this->getSlice($extraRowStart, $extraRowEnd);
                 if (!preg_match('#<w:vMerge/>#', $tmpXmlRow) &&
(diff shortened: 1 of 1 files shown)


vincecore:
Still a problem.

src/PhpWord/TemplateProcessor.php
```diff
@@ -204,7 +204,8 @@ public function cloneRow($search, $numberOfClones)
                 // If tmpXmlRow doesn't contain continue, this row is no longer part of the spanned row.
                 $tmpXmlRow = $this->getSlice($extraRowStart, $extraRowEnd);
                 if (!preg_match('#<w:vMerge/>#', $tmpXmlRow) &&
-                    !preg_match('#<w:vMerge w:val="continue" />#', $tmpXmlRow)) {
```
troosan: It would be better to do only one check
```php
preg_match('#<w:vMerge w:val="continue"\s*/>#', $tmpXmlRow)
```

Issue #341, opened 2014-08-20
deleteBlock / replaceBlock doesn't work properly


timothe:
Hi,

PHPWord is a terrific solution for my needs. Unfortunately I'm not able to use the "deleteBlock" feature since PHPWord doesn't detect the blocks in my files.
I raised a ticket in StackOverflow: http://stackoverflow.com/questions/25402045/regexp-in-a-word-xml-why-does-it-not-match
Basically I achieved to understand the regexp itself doesn't detect the block in the docx.
Any help would be greatly appreciated.

Cheers

judgej:
OMG - this thing uses regex to parse XML!?! I was searching for why my `cloneBlock()` suddenly stopped working after adding some formatting inside the block, and came across this issue. XML and REGEX? Makes the hairs on my neck stand up.

And looking at my document source, yes, it is the REGEX that no longer matches. MS Word puts in a few additional elements of formatting, and the the block tags can no longer be found.

samdark:
Duplicate of #316

klado:
I'm still having issue when cloneBlock and replaceBlock don't find anything (regex tester shows catastrophic backtracking). 

**brad-jones**'s solution solves my problems

michakpl:
@klado probably you already solved that problem, but for other users which find this issue in future I found solution, preg_match used here have limit in PHP, and when document is long it can find only last block, not that one on the beginning of the document.

You have to increase that limit in php.ini file, for bigger files event limit suggested in my solution can be too small:
```
pcre.backtrack_limit = 53001337
pcre.recursion_limit = 53001337
```

Gadeoli:
Hi,

On version 0.16.0 for me i have problems with replace block and delete block (the template gets broken).

For now i just changed the replaceBlock method.

````
public function replaceBlock($blockname, $replacement)
    {
        preg_match(
            //removed (<\?xml.*) from beginning regex
            '/(<w:p.*>\${' . $blockname . '}<\/w:.*?p>)(.*)(<w:p.*\${\/' . $blockname . '}<\/w:.*?p>)/is',
            $this->tempDocumentMainPart,
            $matches
        );

        if (isset($matches[3])) {
            $this->tempDocumentMainPart = str_replace(
                $matches[2] . $matches[3] . $matches[4],
                $replacement,
                $this->tempDocumentMainPart
            );
        }

        //added to remove the start block sign
        $this->setValue($blockname, '');
    }
````

dva-re:
[...]

PR #541, opened 2015-06-01
Update cloneBlock and replaceBlock

chervaliery:
Implements the fonction gave by @brad-jones on the issue :
https://github.com/PHPOffice/PHPWord/issues/341#issuecomment-63755727

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -233,22 +233,25 @@ class TemplateProcessor
     public function cloneBlock($blockname, $clones = 1, $replace = true)
     {
         $xmlBlock = null;
-        preg_match(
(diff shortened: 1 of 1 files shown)


Kertudex:
Hello, 

Do we have a chance to get the PR merged ? What is needed to go ahead  ?

sivaram-blr:
This fix helped me lot. But not understanding why this is not yet merged.

FBnil:
Love the code, it uses SimpleXMLElement/xpath however, how does findBlock do it when there are multiple blocks named the same? would it get the first open and last close block?

Also, why not implement it as a subclass of TemplateProcessor, this way, it can live side-by-side with the original TemplateProcessor.

You also need some testcases to make the code maintainable later on. You should update your code to force Scrutinizer/Travis do a recheck. As it is now, red, you make less chance to merge it in.

Issue #569, opened 2015-07-06
addFontStyle() not working as expected?


iget-master:
Just following PHPWord Sample01 (https://github.com/PHPOffice/PHPWord/blob/develop/samples/Sample_01_SimpleText.php)
and output to HTML file

And it simply don't format my text correcty using "addFontStyle".

Output

```html
<!DOCTYPE html>
  <!-- Generated by PHPWord -->
  <html>
  <head>
  <meta charset="UTF-8" />
  <title>PHPWord</title>
  <style>
  * {font-family: Arial; font-size: 10pt;}
  a.NoteRef {text-decoration: none;}
  hr {height: 1px; padding: 0; margin: 1em 0; border: 0; border-top: 1px solid #CCC;}
  table {border: 1px solid black; border-spacing: 0px; width: 100%;}
  td {border: 1px solid black;}
  .rStyle {font-size: 16pt; font-weight: bold; font-style: italic; text-decoration: line-through ;}
  .pStyle {margin-bottom: 5pt;}
  h1 {font-weight: bold;}
  </style>
  </head>
  <body>
  <h1>Welcome to PhpWord</h1>
  <p>Hello World!</p>
  <p>&nbsp;</p>
  <p>&nbsp;</p>
  <p>I am styled by a font style definition.</p>
  <p>I am styled by a paragraph style definition.</p>
  <p>I am styled by both font and paragraph style.</p>
  </body>
  </html>
```

Using latest PHPWord version get from Composer.

RKaczmarek:
already discussed in #556

iget-master:
Created PR #570, now everything works fine! Please merge it! :+1:

stevenmunro:
Is this still an issue? I'm confused. addFontStyle doesn't appear to be working with the latest release either.

PR #570, opened 2015-07-07
Fix bug with text style at HTML Writer

iget-master:
This PR fixes the issue #569 created by me.

Both getFontStyle() and getParagraphStyle() was ignoring style if it was not an style object.

Now if it's not an object, the style is the class name to be applied to the element.

### DIFF
--- a/src/PhpWord/Writer/HTML/Element/Text.php
+++ b/src/PhpWord/Writer/HTML/Element/Text.php
@@ -146,17 +146,19 @@ class Text extends AbstractElement
     {
         /** @var \PhpOffice\PhpWord\Element\Text $element Type hint */
         $element = $this->element;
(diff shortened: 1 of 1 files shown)


iget-master:
@RomanSyroeshko Please take a look at this, it's an production bug for me. :(

iget-master:
Could someone take a look here?

troosan:
@iget-master could you please check whether this still occurs on the dev-develop branch? I should not be a problem anymore.

iget-master:
We have migrated to phpdocx project, so I cannot confirm this bug anymore. :-(

troosan:
indeed, it took 2 years to have it fixed. I hope I'll manage to roll out more regular releases!

Issue #574, opened 2015-07-16
Incorrect Constant Names in \Style\Font.php


ComputerTinker:
I was just looking through the code and noticed that lines 36 and 37 which currently read:

``` php
    const UNDERLINE_DOTHASH = 'dotDash';
    const UNDERLINE_DOTHASHHEAVY = 'dotDashHeavy';
```

...should in fact read (notice HASH vs DASH in the constant name):

``` php
    const UNDERLINE_DOTDASH = 'dotDash';
    const UNDERLINE_DOTDASHHEAVY = 'dotDashHeavy';
```

Progi1984:
Fixed in #578 by @troosan

PR #578, opened 2015-07-22
 #574 Incorrect Constant Names in \Style\Font.php

ComputerTinker:
Correct constants mapped to "dotDash" and "dotDashHeavy", leaving old spellings for backwards compatibility.


ComputerTinker:
Just looked at the Travis CI build error and it does not seem to be a problem with my code:

[Composer\Repository\RepositorySecurityException]                            
  The contents of http://packagist.org/p/provider-2015-04$eb238dbb58947a5f32a  
  394f4f4280136fed9a157da92403980f51e0d6bd06a40.json do not match its signatu  
  re. This should indicate a man-in-the-middle attack. Try running composer a  
  gain and report this if you think it is a mistake.

PR #579, opened 2015-07-24
Expose 'through' and 'topAndBottom' Image Wrapping Styles

ComputerTinker:
The 'through' and 'topAndBottom' image wrapping styles already exist in \Styles\Frame.php. Just exposing them in the documentation and samples, and adding some backwards compatible constants.


troosan:
Those values are there only for backward compatibility, not to break any code that is using it.
The values in Image will eventually be removed.
Image::WRAP_THROUGH and Image::WRAP_TOPBOTTOM should be used instead.

PR #581, opened 2015-07-29
Update Html.php

SSchwaiger:
I'm import HTML generated by Froala, a WYSIWYG JavaScript editor. IT generates

tags in each

item. As a quickwin without having to parse all child tags (which is for sure a new add on as well to be able to format list items!), I just add the raw node content.


troosan:
same fix was done in commit ac357d10d52b8c2e491e56683f8a07665d3fa5ff

PR #585, opened 2015-07-31
Fix FormField Constructor Ignoring Styles

ComputerTinker:
I noticed that the font and paragraph styles I supplied when adding a
new FormField were ignored. It seems that the FormField class extends
the Text class and overwrote the constructor, but the code to set the
paragraph and font styles was mistakenly omitted. I just copied and
pasted what was in the Text class constructor, and now it works like a
charm!


troosan:
did the same change, but calling the super constructor instead

PR #586, opened 2015-08-01
addHtml() support of _b_ and _i_ tags

maximeplante:
I have noticed that addHtml supports the _em_ and _strong_ tags but not the _b_ and _i_ tags.

Do you think we should support these tags even if their use as bold and italic is [now considered a bad practice](http://www.w3.org/International/questions/qa-b-and-i-tags) ?


troosan:
already implemented in dev branch

Issue #587, opened 2015-08-03
setAlign() vs. setAlignment()


ComputerTinker:
I notice that the Frame, NumberingLevel, Table, and Paragraph elements can all be aligned.  For the NumberingLevel element the attribute is named 'align', and the methods used are getAlign() and setAlign().  So far so good. For the Paragraph element the attribute is named 'alignment', and the methods used are getAlignment() and setAlignment(). That works for me as well.  

Where the code seems inconsistent is with the Frame and Table elements.  There the attributes are named 'alignment', but the methods are getAlign() and setAlign().  I would have expected the methods to be getAlignment() and setAlignment() so that they reflect the attribute name, like most (all?) of the other methods I've seen so far do.  Not a big deal, just a minor nitpick about consistency.  Thanks.

RomanSyroeshko:
All the fields as well as correspondent accessors and mutators will be renamed in ``0.13.0``. ``align`` will be replaced with ``alignment``, ``setAlign``/``getAlign`` with ``setAlignment``/``getAlignment``.

RomanSyroeshko:
[Follow PHPWord](https://twitter.com/PHPWord/) on Twitter to catch the moment when we are ready for alpha testing of ``0.13.0`` release. I would like to make the product more stable, so we will need some volunteers soon. :smirk:

PR #588, opened 2015-08-03
#587 Add getAlignment() setAlignment to Frame and Table

ComputerTinker:
Add getAlignment() and setAlignment methods to the Frame and Table
elements, so that the method names are consistent with the attribute
names.  Preserve the old methods for backwards compatibility.  (Issue #587)


RomanSyroeshko:
I'm working on that already.

Issue #349, opened 2014-08-25
Table width is documented as percentage but needs to be multiplied with 50 for DOCX


gmta:
The documentation about the 'table' element states:

```
Table styles:
-  ``width`` Table width in percent
```

However, if you want to have a 100% width table in a DOCX file, you need to multiply the value with 50 (e.g. 100% = 100 \* 50 = 5000). The same goes for cell width.

jelofson:
Agreed. This should be documented as it is very confusing.

ComputerTinker:
This issue snagged me as well. Just created a patch and submitted a pull request for it.

rebekahocampo:
Does anyone know the status of this bug? I just installed phpoffice/phpword a couple days ago, and while developing my code, I noticed this same thing is happening. Very Confusing.

sonalithakkar:
include("PHPWord/bootstrap.php");
    include_once('Sample_Header.php');
   $phpWord = new \PhpOffice\PhpWord\PhpWord();
   $section = $phpWord->addSection();
    $html = '<table cellpadding="0" cellspacing="0" border="0" id="invoice_content" style="width:100%" class="table table-bordered">
            <tr>
                <th style="text-align: center;">INVOICE</th>
            </tr>
            <tr>
                <td width="100%">
                    <table cellpadding="0" cellspacing="0" border="0" class="table table-bordered" align="center" aria-describedby="example_info" width="100%">
                        <thead>
                            <tr>
                                <th style="text-align:left;">Account No.</th>
                                <th style="text-align:left;">Invoice Number</th>
                            </tr>
                        </thead>
                    </table>
                </td>
            </tr>
        </table>'; 
\PhpOffice\PhpWord\Shared\Html::addHtml($section, $html, false, false);

// Save file
write($phpWord, $report_name, $writers,$path.$report_name);
if (!CLI) {
     include_once 'Sample_Footer.php';
 }

In this width of table not taking 100%.
Table not showing 100% to doc file

If there are more <th> then it's taking full length of doc.

The output of parseStyle is and node name also printed

#text=>style=>body=>table=>

Array
(
    [width] => 5000
    [unit] => pct
)
tr=>th=>tr=>table=>

Array
(
    [width] => 5000
    [unit] => pct
)
body=>th=>#text=>th=>tr=>table=>

please help ASAP

PR #589, opened 2015-08-03
#349 Documentation Updates: Table with Percentage and Alignment

ComputerTinker:
Note that table with percentage is actually specified in 1/50ths of a percentage (issue #349), and add note about table alignment as well.


troosan:
documentation has been reorganised, no applicable anymore

PR #594, opened 2015-08-12
Word 2007 Reader: Added support for ListItemRun

atomicalnet:
Some data is lost while reading lists with Word2007 parser.
Lists items are parsed as ListItem elements, this class is not meant to be a container, therefore if a list item contains a link or an image it will be ignored.

This fix wraps all child elements of a list item inside a ListItemRun, a more suitable class which is a container and that will allow us to hold, text, links, images, etc. inside as elements.


Azuka:
:+1:

troosan:
@atomicalnet your branch is read-only so I'm continuing this on a new one (#1290)

Issue #526, opened 2015-04-23
addTextBreak causes Error


MarcTroesken:
If I try to add a new text break with:
```php
$section->addTextBreak('1');
```

This errors shows up:

```php
ErrorException in AbstractContainer.php line 101:
Undefined offset: 2
```

any ideas on that ?

ComputerTinker:
I suspect you're confusing it by passing in a string instead of a number.  Try this instead:

```php
$section->addTextBreak(1);
```

KvanSteijn:
I have the same error by using this code:
```php
$section->addTextBreak(1);
```

ComputerTinker:
Interesting. Line 101 of AbstractContainer.php reads:

```PHP
@list($count, $fontStyle, $paragraphStyle) = $args; // Suppress error
```

The '@' is supposed to suppress errors generated by PHP's list function, but there's a note on this page http://php.net/manual/en/language.operators.errorcontrol.php which says "If you have set your own default error handler then the error still gets sent to the error handler regardless of the @ sign."

I wrote a quick little test script.  When I just use the following, no error is reported:

```PHP
$objPHPWord = new \PhpOffice\PhpWord\PhpWord();
$section = $objPHPWord->addSection();
$section->addTextBreak(1);
```

But if I add a custom error handling, I get two errors.  The first one fires because the font style is not set, and the second one fires because the paragraph style is not set:

```PHP
function myErrorHandler($errno, $errstr, $errfile, $errline) {
  echo "I hit an error!";
}
set_error_handler("myErrorHandler");

$objPHPWord = new \PhpOffice\PhpWord\PhpWord();
$section = $objPHPWord->addSection();
$section->addTextBreak(1);
```

If you wish to keep your custom error handler in place and just work around the issue, you can specify all your text breaks like this:

```PHP
$section->addTextBreak(1, null, null);
```

If you wish to patch your code so that you don't need to add the extra nulls each time you add a text break, I believe replacing line 101 in AbstractContainer with the following will do the trick:

```PHP
$count = args[0];
if (isset(args[1]))
  $fontStyle = args[1];
if (isset(args[2]))
  $paragraphStyle = args[2];
```

ComputerTinker:
I just created a pull request for this issue. You may also feel free to download from https://github.com/ComputerTinker/PHPWord, which is up to date with the latest PHPWord as of today, plus has a few other patches applied.

PR #607, opened 2015-08-31
Fix Error Suppression in Abstract Container

ComputerTinker:
When the developer user has a custom error handler, the @list syntax
does not work. Consequently, when addTextBreak() is called with only one
parameter, it throws errors. Refactor to prevent errors from being
thrown in that situation.  Fixes issue #526.


KvanSteijn:
This pull request fixed my problem, but is see that this pull request is the same pull request as  https://github.com/PHPOffice/PHPWord/pull/469

ComputerTinker:
You are right, my fix is a duplicate of the previous one. Sorry about that, I completely missed the earlier PR.

PR #624, opened 2015-09-22
Do Not Force Table Cells to Have Widths

ComputerTinker:
Previously, all table cells had widths which resulted in Word
interpreting the cell as having a preferred width of zero, which
squished the cell contents.  This change does not write out the width
attribute when the width is zero, which then allows the cell's width to
truly be dynamic based on the contents of the cell.


troosan:
similar change was already merged

PR #638, opened 2015-10-16
Updated composer installation instructions

janhohner:
It is now recommended to use the command line to add new dependencies to a project.


ssanders:
:+1:

PR #660, opened 2015-11-16
renamed PhpWord/Shared/String to avoid conflicts

Haringat:
String is a reserved classname


Haringat:
This can be closed since #665 does the same.

PR #658, opened 2015-11-08
Added PhpOffice\Common as requirement

Progi1984:
@RomanSyroeshko : I wait for your validation :)

PR #665, opened 2015-11-20
Bump to travis php version

michaelachrisco:
Small code coverage on section footers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,12 @@ php:
     - 5.4
     - 5.5
     - 5.6
+    - 7.0
     - hhvm
 
 matrix:
     allow_failures:
+        - php: 7.0
         - php: hhvm
 
 env:
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "phpmd/phpmd": "2.*",
         "phploc/phploc": "2.*",
         "dompdf/dompdf":"0.6.*",
-        "tecnick.com/tcpdf": "6.*",
+        "tecnickcom/tcpdf": "6.*",
         "mpdf/mpdf": "5.*",
         "zendframework/zend-validator": "2.5.*"
     },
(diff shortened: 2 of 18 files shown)


michaelachrisco:
Looks like String is a protected word in php 7.0. Ill add in a patch (String -> PHPWordString).

michaelachrisco:
Two bugs:
* Methods with the same name as their class will not be constructors in a future version of PHP; PclZip has a deprecated constructor

* Methods with the same name as their class will not be constructors in a future version of PHP; mPDF has a deprecated constructor

Looks like either these packages need to be updated or fixed of PHP 7 errors. Down the rabbit hole I go.

Haringat:
You should revert 72927e3.
The package tecnick.com/tcpdf is abandoned and replaced by tecnickcom/tcpdf so the former dependency list was correct.

michaelachrisco:
@Haringat Ill take a stab at this later today. I have a feeling that it wont be as simple as requiring the correct library and off we go. I have no idea if the two libraries can work with PHP 7.0. tecnickcom/tcpdf Does not have a travis build by the look of it

michaelachrisco:
I believe I am going to let 7.0 fail at the moment (unless someone else has a better option?), and make the test pass. I'll go down the rabbit hole on some of the dependencies to fix the 7.0 issue on their side and then push to this branch.

RomanSyroeshko:
Partially duplicate of #658.

As regards coverage. I made couple experiments and found that `@codeCoverageIgnore` annotation sometimes has no effect. I have no idea why. For example, you can see that `\PhpOffice\PhpWord\Element\Footnote` has four methods, two of which are marked with `@codeCoverageIgnore`. In the same time the report says that coverage is 50% (2 of 4 methods are covered). That looks strange, because two methods should be ignored at all.

I think we should consider upgrading PHPUnit in the future. We use very old one, and it may have some bugs. Thanks for pointing this out.

michaelachrisco:
Thanks @RomanSyroeshko I was stuck for a while on some of the tests and why they were not passing. Going into the parent repos was a rabbit hole. Glad to see someone working on 7.0 coverage.

PR #666, opened 2015-11-24
Update AbstractContainer.php

kaigoh:
Even though the @list is trying to suppress the PHP error, it still appears in my CodeIgniter logs. My fix gets rid of this error.


troosan:
already fixed

PR #668, opened 2015-11-27
patched html reader to read HTML (not only XML) documents

xprt64:
The current reader (PhpOffice\PhpWord\Shared\HTML) breaks if the HTML document is not XHTML (because of the call to \DOMDocument::loadXML).

As the name suggests, it should parse HTML: the class is named 'Html' and the static method is named 'addHtml'.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ vendor
 phpword.ini
 /.buildpath
 /.project
+/nbproject/private/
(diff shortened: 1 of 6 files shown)


troosan:
@xprt64 Thanks for this interesting PR, but I don't want to add a dependency to a dev version of an external library.

PR #679, opened 2015-12-04
Xprt64

xprt64:
- Improved HTML parser (+css parsing/styling)
- BUG: PHP_INT_MAX is too large for z-index of images


troosan:
see #668

Issue #355, opened 2014-08-30
Source code migration to 5.4 features


ghost:
PHP developers [announced](http://php.net/archive/2014.php#id2014-08-14-1) end of life of PHP 5.3. Since now the minimal supported version is 5.4. We should plan migration to PHP 5.4.

Migration guide from PHP 5.3.x to PHP 5.4.x could be found [here](http://php.net/manual/en/migration54.php).

Progi1984:
PHPWord is not anymore compatible PHP 5.4.

PR #698, opened 2016-01-08
PHP7 Update - Renamed String class to SharedString

publiux:
PHP7 now reserves the class name String. PHPWord utilizes String as a class name, thus PHPWord presently does not work with PHP7. This PR renames the class to SharedString and updates it usage through the library and in the tests.


publiux:
This build is failing because of mpdf.  mpdf is still using the old constructor style that is not suitable for PHP7. A fix was pulled into mpdf's development branch ( https://github.com/mpdf/mpdf/issues/68 ) but has not yet been tagged and thus is not pulled in via composer package.

publiux:
I have updated composer.json to pull dev-development for mpdf, which has merged the constructor fix for PHP7. In the future, once the fix is pulled into a release tag, we need to change the composer.json to a stable version. Hopefully, this won't take too long.

RomanSyroeshko:
Partially resolved in scope of #658.

Faryshta:
what does this need to be merged?

RomanSyroeshko:
Some thoughts.

mPDF 6.1 requires PHP 5.4+ while PHPWord - PHP 5.3.3+.

So, first of all #355 should be resolved.

Issue #695, opened 2015-12-31
Unable to read .doc files


mayankitt:
Hello,
I am trying to load 97-2003 MS Word documents but the following error keeps showing up.

![image](https://cloud.githubusercontent.com/assets/11437102/12063099/dc6ce944-afce-11e5-9809-1c969498f832.png)

Following is simply an adaptation of the code provided in the [Sample_11_ReadWord97.php](https://github.com/PHPOffice/PHPWord/blob/develop/samples/Sample_11_ReadWord97.php) file:

```php
        $source = "OldDoc.doc";
        echo "Reading document ".$source."<br><br>";
        if(str_contains($source, '.docx'))
            $phpWordObj = IOFactory::load($source);
        elseif(str_contains($source, '.doc'))
            $phpWordObj = IOFactory::load($source, 'MsDoc');
        echo "Document reading complete";
```
I am using Laravel 5.1 and docx files seem to load just fine. Any idea where I might be going wrong with this?

vovanada:
Library has bug in this place https://github.com/PHPOffice/PHPWord/blob/v0.12.1/src/PhpWord/Reader/MsDoc.php#L144

class OLERead hasn't property documentSummaryInformation. It rename in this commit https://github.com/PHPOffice/PHPWord/commit/90d64bd3475c05814d47ba497dba5c392308b7cd

You have four way to solve this problem:

1. Wait for fix this library by developers
2. Fix it
3. Set error_reporting to 0
4. Use old version of library

Similar topic:

https://github.com/PHPOffice/PHPWord/issues/610
https://github.com/PHPOffice/PHPWord/issues/503

RomanSyroeshko:
See #503.

PR #612, opened 2015-09-02
fix MsDoc reader

jaberu:
Fixed issue: Undefined property: PhpOffice\PhpWord\Shared\OLERead::$documentSummaryInformation

https://github.com/PHPOffice/PHPWord/issues/610

PR #701, opened 2016-01-12
fx issue #695

vovanada:
fix reader for word97. It work only with error_reporting 0;


troosan:
already fixed and merged in #612

Issue #548, opened 2015-06-09
TemplateProcessor, some values are not replaced (incl. debug)


kaystrobach:
Given you have word document, with placeholders like ```${some.object.path}``` any change in any place of the document can add some xml inside the brackets of the placeholder.

```
$template = new \PhpOffice\PhpWord\TemplateProcessor($this->templatePath);
throw  new \Exception(print_r($template->getVariables(), TRUE));
```

gives f.e. the following array

```
[0] => student.aktenzeichen
[1] => student.registraturSba
[2] => student.kontakt.vorname
[3] => st</w:t></w:r><w:r><w:t>u</w:t></w:r><w:r><w:t>dent.kontakt.nachname
[4] => student.kontakt.strasse
[5] => st</w:t></w:r><w:r><w:t>u</w:t></w:r><w:r><w:t>dent.kontakt.hausnummer
[6] => student.kontakt.plz
[7] => student.kontakt.ort
[8] => student.kontakt.nachname
```

```3``` and ```5``` will never be replaced as they are screwed up by word :cry: as i use getVariables to obtain real object pathes from my domain model i used strip_tags to transform it into a useable object path again.

But neither giving the broken values ```3``` / ```5``` to set values gives a successfull replacement ... any idea? my current approach is to digg deeper into the ```setValueForPart``` function and understand what you do with value cleaning ...

**Comment by kaystrobach** [comment]
the following workaround makes the replacement a lot more robust for me!

current workaround:

```
		$variables = $template->getVariables();

		foreach ($variables as $variable) {
			if (strpos($variable, '.')) {
				list($object, $path) = explode('.', strip_tags($variable), 2);
				if (isset($this->variables[$object])) {
					$value = $this->resolve($this->variables[$object], $path);
					$template->setValue($variable, $value);
					$template->setValue(strip_tags($variable), $value);

					$this->logger->log('WordView Variable: ' . $variable . ' = ' . $value, LOG_DEBUG);
				}
			}
		}

		$this->logger->log('WordView leftovers', LOG_DEBUG, print_r($template->getVariables(), TRUE));
```

mean i do 2 replacements per document, one with the xml tags included and one with the xml removed with strip tags ... this makes it a little more robust for me, IMHO this should be fixed with some kind of mapping table, as the correct areas are detected.

RomanSyroeshko:
Fixed.

PR #705, opened 2016-01-17
Bugfix #548: variables edited in word get messed with invisible XML

d-damien:
Hi,

So in our experience with Word templates, whenever someone edits a variable name without rewriting it entirely (the whole "${variable}"), it becomes a mess with invisible XML, preventing the variable to be replaced with intended content later.

I wrote this to clean Word templates variables before calling ``setValue()``. It should simply be used this way :
```php
$template = new TemplateProcessor(...);
$template->fixBrokenVariables();
$template->setValue(...);
// etc.
```
I was not sure wether I should include that in the constructor (a document *should* be clean by default I guess), but since it could be too much on a lot of documents maybe (?), I made a separate function so you can call it when you need it.

What do you think, @kaystrobach ?

(I'm new to Github, tell me if I did something wrong.)

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -136,6 +136,43 @@ class TemplateProcessor
     }
 
     /**
+     * Clean all template parts of hidden XML within variables/placeholders.
+     *
+     * @return void
+     */
+    public function fixBrokenVariables()
+    {
+        $this->tempDocumentMainPart = $this->fixBrokenVariablesForPart($this->tempDocumentMainPart);
+
+        foreach ($this->tempDocumentHeaders as &$headerXML) {
+            $headerXML = $this->fixBrokenVariablesForPart($headerXML);
+        }
+
+        foreach ($this->tempDocumentFooters as &$footerXML) {
+            $headerXML = $this->fixBrokenVariablesForPart($footerXML);
+        }
+    }
+
+    /**
+     * Fix dirty hidden XML within variables in document part.
+     *
+     * @param string $documentPartXML said document part
+     *
+     * @return string
+     */
+    protected function fixBrokenVariablesForPart(string $documentPartXML)
+    {
+        return
+            preg_replace_callback(
+                '|\$([^{]*)(\{.*})|U',
+                function ($match) {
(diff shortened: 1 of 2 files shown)


publiux:
This is a great bugfix. However, the test case is failing due to insufficient variables provided to the TemplateProcessor constructor. Fix that in your forkand you should be good to go. I'm anxious to have this pulled. Thanks.

kaystrobach:
will test it out, could you please fix the test :) and i would prefer to fix the variables by default in the constructor and only have an option to avoid that.

Here is why:
Most developers do not want to digg deep in the output formats, they just want to get it working, and so the library should be as intelligent as possible and as we know the problem we should fix it predictable.

publiux:
@kaystrobach The only thing missing from the constructor I mentioned above is a test word file. You should create one with the garbled XML and then what the final output should be. In your test case, pass in the path to the garbled XML Word Document and assert that the final result is equal to the "good" Word document you make. That will get the test passing.

d-damien:
@publiux Yes, that's what I was thinking. Sorry for the delay ; I'm gonna take care of it.

publiux:
@RomanSyroeshko Any chance of accepting this PR soon?

RomanSyroeshko:
Hi guys,
The problem with broken macros was resolved long time ago (see `TemplateProcessor::fixBrokenMacros`).

If there are any issues with that now, please, provide me with reproducible test cases.

Thanks!

publiux:
@RomanSyroeshko I'm not sure this has to do with broken macros. This PR seems to fix the issue where saving a Word file after you have created it with the ${} variable substitution format adds XML elements between some of those (${}) characters. Thus, if you create the template, edit it, and then run it through PHPWord, in some instances, the variable will not be replaced because PHPWord can no longer find the substitution sequence. That said, I tested this on my own project and this does not work correctly. A new fix is required.

RomanSyroeshko:
Yes, I mentioned exactly the same problem you've described. If you have any issues with that, please, provide step by step instruction to reproduce your case. Thanks!

RomanSyroeshko:
https://github.com/PHPOffice/PHPWord/issues/548#issuecomment-207823195.

Many thanks for the test. We definitely missed one case.

PR #721, opened 2016-02-02
Update styles.rst

jeffersoncarpenter:
I had guessed "solid" but ended up reading the PHPWord code to find out "single".

Right here https://phpword.readthedocs.org/en/latest/elements.html#font-style is actually the documentation I was reading at the time; I would make the change there too.


jeffersoncarpenter:
should i do this on the master branch?

Issue #720, opened 2016-02-02
Cannot read and immediately write back out Word file with invalid XML entities in content


tjarrett:
If you load a Word 2007 file that has invalid XML characters in the content (for example [xml-characters-in.docx](https://github.com/PHPOffice/PHPWord/files/114558/xml-characters-in.docx)) and then immediately try to save that file back out doing something like this:

```
$test = \PhpOffice\PhpWord\IOFactory::load('xml-characters-in.docx');
$test->save('xml-characters-out.docx');
```
You get a corrupted Word file back (see [xml-characters-out.docx](https://github.com/PHPOffice/PHPWord/files/114559/xml-characters-out.docx)).

This is probably related to #671, #401, #514, and other similar issues. However, I don't have a chance to scrub the content because I am reading and then immediately trying to save the result. 

Maybe the Reader should scrub the incoming $textContent?

EasySoftwarePoland:
hello, any news on it? I've got the same problem...

PR #722, opened 2016-02-02
Escape invalid XML characters when reading in a Word2007 file

tjarrett:
This fix resolves issue #720 in which reading in and then immediately saving a Word2007 file fails if there are any invalid XML characters in the document.

I realize it is the user's responsibility to escape content when using the Writer (and actually agree with this stance), but it feels like it is this library's responsibility to do so when reading in a document.

To test this I fed [xml-characters-in.docx](https://github.com/PHPOffice/PHPWord/files/114710/xml-characters-in.docx) into PHPWord using this code:

```
$test = \PhpOffice\PhpWord\IOFactory::load('xml-characters-in.docx');
$test->save('xml-characters-out.docx');
```

And the result is the attached file [xml-characters-out.docx](https://github.com/PHPOffice/PHPWord/files/114709/xml-characters-out.docx).


troosan:
merged manually

PR #416, opened 2014-11-10
Properly fixing #257 and helps out on #324

twmobius:
I think this properly solves #257 replacing `method_exists` with `is_callable` as well as empty tables being generated.

Issue #324, opened 2014-08-01
Using table tag in HTML Reader produces no output


EK1771:
**Sample Code:**

```
$phpWord = new \PhpOffice\PhpWord\PhpWord();

$section = $phpWord->addSection();

$html = '<table><tr><td>test</td></tr></table>';

\PhpOffice\PhpWord\Shared\Html::addHtml($section, $html);

$objWriter = \PhpOffice\PhpWord\IOFactory::createWriter($phpWord, 'Word2007');
$objWriter->save('test.docx');
```

**Expected Output:**

Table with one cell containing the word "test".

**Actual Output:**

Blank

From stepping through the code quickly, the issue seems to be caused by the following if condition in parseChildNodes():

```
if ($element instanceof AbstractContainer) {
    self::parseNode($cNode, $element, $styles, $data);
}
```

Commenting out the if condition then allows for the sample code above to produce the expected output.

matteomoretti:
Any news? The bug still occours

phpnets:
by using \PhpOffice\PhpWord\Shared\Html::addHtml($section, $html)  we can interpret html to word. Can we set alignment options for this output (such as align right/left/both) ?

garethellis36:
@mogilvie are you able to share your complete and working `Html` class with modifications? I'm trying it myself but when I try and apply it to an updated sample (as below), I get an error when it tries to write because of the objects is null.

```php
 $html .= '<table><thead><tr><th>Header of column 1</th><th>Header of column 2</th></tr></thead>';
$html .= '<tbody><tr><td>Row 1 for column 1</td><td>Row 1 for column 2</td></tr></tbody></table>';
```

mogilvie:
I'll post the full class tonight, but in the meantime, is the error caused by the switch statement not having any code to be executed for case: 'skiphead'?

garethellis36:
@mogilvie I doubt that - it should fall through to the 'skipTbody' case because there's no break statement. I don't fully understand how this works but I *assumed* that thead would need to be handled in the same way as tbody.

mogilvie:
@garethellis36 Agree re skipThead. Is the parseTable function returning the $newElement? It's cut off from the sample html class.

mogilvie:
Html.php uploaded as a text file. It was working as of Jan 2015.

[Html.txt](https://github.com/PHPOffice/PHPWord/files/119668/Html.txt)

surindersinghva:
[...]

PR #724, opened 2016-02-04
Resolves certain HTML tags rendering no output

### LINKED ISSUE #257: <p> and header tags puts no output
Even in Sample_26_HTML no output seen in docx file for the parts <b>p</b> and <b>h1</b>. Those tags are always giving empty output

garethellis36:
This is a continuation of PR #416 which appears to have been abandoned by @twmobius. The commits herein are just the same commits that he submitted, but I have addressed the concerns raised by @Progi1984 in [this comment](https://github.com/PHPOffice/PHPWord/pull/416#issuecomment-66249254).

This PR should resolve #257 and #324.

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -177,9 +177,7 @@ class Html
             $cNodes = $node->childNodes;
             if (count($cNodes) > 0) {
                 foreach ($cNodes as $cNode) {
-                    if ($element instanceof AbstractContainer) {
-                        self::parseNode($cNode, $element, $styles, $data);
-                    }
+                    self::parseNode($cNode, $element, $styles, $data);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


twmobius:
It's good that someone took the time to fix this. I am sorry I haven't been able to get more involved. 
:+1:

derimagia:
Great I found this issue as well and fixed it but was surprised no one has had issues with it.

However, the fix should be replaced with this:
```php
if ($element instanceof \PhpOffice\PhpWord\Element\AbstractElement) {
    self::parseNode($cNode, $element, $styles, $data);
}
```

The instance didn't have a FQN attached to it. Here's someone explaining the problem: http://php.net/manual/en/language.operators.type.php#106882

src/PhpWord/Shared/Html.php
```diff
@@ -177,9 +177,7 @@ private static function parseChildNodes($node, $element, $styles, $data)
             $cNodes = $node->childNodes;
             if (count($cNodes) > 0) {
                 foreach ($cNodes as $cNode) {
-                    if ($element instanceof AbstractContainer) {
```
derimagia: Since this was removed the use statement can be removed at the top I believe

ROunofF:
Is this are reason not to merge this ? (Like breaking anything?) It seems to do the job...

PR #727, opened 2016-02-12
Add TextRun as container for CheckBox

irond:
This fixes exception if you want to put checkbox inside textrun which is inside a table cell. See code above.

``` php
$table = $section->addTable();
$table->addRow();
$cell = $table->addCell(2000);
$textRun = $cell->addTextRun();
$textRun->addCheckBox('test', ' yes');
$textRun->addCheckBox('test', ' no');
```


troosan:
No sure why the FormField is allowed and not the CheckBox ...
I guess the CheckBox should be obsoleted in favour of the FormField.

PR #759, opened 2016-03-15
Add method to change default state of checkbox

irond:
You can now set default state of checkbox using method setChecked()

``` php
$section->addCheckBox('test1', '')->setChecked('true'); //checked
$section->addCheckBox('test1', '')->setChecked('0'); //unchecked
```

### DIFF
--- a/src/PhpWord/Element/CheckBox.php
+++ b/src/PhpWord/Element/CheckBox.php
@@ -32,6 +32,12 @@ class CheckBox extends Text
      * @var string
      */
     private $name;
+    
+    /**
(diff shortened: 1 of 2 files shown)


troosan:
@irond can you please
- [ ] add a unit test
- [ ] make sure all the code is covered by it
- [ ] update the documentation

troosan:
Actually, if I'm not mistaken this can be achieved by using a FormField of type checkbox instead of using the CheckBox class. I guess this class should be obsoleted in favour of the FormField which supports more types of fields.

PR #763, opened 2016-03-20
Clarifying the require_once line when using Composer

wasimj:


troosan:
not applicable anymore

PR #776, opened 2016-04-07
Add missing Style getters & setters for Element classes.

samsullivan:
In my project, I extended PHPWord's functionality to create universal "paragraph" objects (AKA elements that accept paragraph styles). I needed to do this so that I could add styles (namely `keepNext`) to the last paragraph of set of paragraphs. When doing this, I noticed that a lot of elements that accepted paragraph styles on creation didn't have getter/setter functions for further updating these styles. See the errors thrown by my IDE below:

screenshot from 2016-04-07 17 56 38

My project currently only uses `Element\Text`, `Element\TextRun`, `Element\Link`, and `Element\ListItem`. Because of this, the following PR only has commits to make sure these 4 classes have the relevant getter/setter functions for all styles. I am trying to get a big update to production ASAP, so this is all I can do for right now (i.e. `Element\CheckBox` is missing still).

Let me know if you want to wait on this until I finish the rest of the classes, since I'd be more than happy too. Also, I tried to follow your method style from `Element\Text` even though there were some things I wanted to tweak (didn't want to go refactor crazy on my first PR).

Thanks for the great library, hope this helps!

### DIFF
--- a/src/PhpWord/Element/Link.php
+++ b/src/PhpWord/Element/Link.php
@@ -80,8 +80,7 @@ class Link extends AbstractElement
     {
         $this->source = String::toUTF8($source);
         $this->text = is_null($text) ? $this->source : String::toUTF8($text);
-        $this->fontStyle = $this->setNewStyle(new Font('text'), $fontStyle);
-        $this->paragraphStyle = $this->setNewStyle(new Paragraph(), $paragraphStyle);
+        $this->setFontStyle($fontStyle, $paragraphStyle);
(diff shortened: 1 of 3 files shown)


troosan:
@samsullivan Indeed, there is no setter, but there is a getter.
All you need to do is:
```php
$myLink = $section->addLink();
$myLink->getParagraphStyle()->set...
```

samsullivan:
Thanks for the response, @troosan.

Issue #667, opened 2015-11-26
Images not rendered correctly on LibreOffice


xprt64:
The images are not rendered correctly (at least on LibreOffice), they get enlarged by a few millimeters and thus get fuzzy. I resolved the issue by modifying the width and height attributes from pixels to points in `PHPWord_Writer_Word2007_Base::_writeImage()` like this:

``` php
$objWriter->writeAttribute('style', 'width:'.(0.75 * $width).'pt;height:'.(0.75*$height).'pt');
```

Environment: Linux version 3.16.0-4-amd64 (debian-kernel@lists.debian.org) (gcc version 4.8.4 (Debian 4.8.4-1) ) #1 SMP Debian 3.16.7-ckt11-1+deb8u6 (2015-11-09)

michaelachrisco:
Which version of LibreOffice are you using? Can you upload a sample doc? I may have ran into this bug before.

xprt64:
Version: 4.3.3.2
Build ID: 430m0(Build:2)

The code I posted on is about an old version of PhpWord but the bug is
still there in the last stable version.

2015-11-27 15:28 GMT+02:00 michael chrisco <notifications@github.com>:

> Which version of LibreOffice are you using? Can you upload a sample doc? I
> may have ran into this bug before.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/PHPOffice/PHPWord/issues/667#issuecomment-160139836>.
>



-- 
Best regards,
Constantin Galbenu,
Tel: +40728247366
Skype: xprt64

michaelachrisco:
Ill take a look on my side when I get a chance. If memory serves, the issue has something to do with the GD extension. http://php.net/manual/en/book.image.php

The offender might have something to do with the image element: https://github.com/PHPOffice/PHPWord/blob/develop/src/PhpWord/Element/Image.php

I know I was testing on MS Word and the stretching was happening.

xprt64:
It's not about GD, the image is fine in the docx file and the dimensions
are correct on the docx/word/document.xml; it's just about LibreOffice
rendering: if the unit is pixel then it get streched as if LibreOffice
ignores the unit and uses points anyway.
The solution that worked for me was to use points as unit in the resulting
docx file.

2015-11-27 15:48 GMT+02:00 michael chrisco <notifications@github.com>:

> Ill take a look on my side when I get a chance. If memory serves, the
> issue has something to do with the GD extension.
[...]

PR #779, opened 2016-04-11
Save PNG alpha information when using remote images.

samsullivan:
PNG images downloaded from remote sources can be either blurry or, worse, very glitched. This is because [`imagesavealpha($image, true)`](http://php.net/manual/en/function.imagesavealpha.php) is not run automatically when PNG images are downloaded, created, and saved locally for file creation.

RE: #667


samsullivan:
Please note, I did not test the change to `Element\Image::getImageStringData()` since my application doesn't use this. If someone wants to run a test, [similar to this](https://github.com/PHPOffice/PHPWord/issues/667#issuecomment-207142448), with an HTML writer, that would be great.

The logic updated in `Writer\AbstractWriter::addFilesToPackage()` does fix the problem with the Word2007 writer.

lalop:
@troosan is it possible to merge this pr to develop branch ?
Thank you ?

PR #784, opened 2016-04-19
TemplateProcessor: add the ability to specify pattern prefix, suffix and case sensitivity

ctrlaltca:
This PR adds some properties and a method to TemplateProcessor that can be used to specify a different prefix and suffix for the search pattern. It still defaults to ${ and }, so it's backwards compatible.
Additionally, the search pattern case sensitivity can be defined; it defaults to true to mimic the current behavior.

This is currently based on master. If this is ok, i'll be glad to port it to the development branch, too.

This is somewhat similar to #346.

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -61,6 +61,27 @@ class TemplateProcessor
     protected $tempDocumentFooters = array();
 
     /**
+     * Search pattern prefix. Defaults to ${
+     *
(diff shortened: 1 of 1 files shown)


ctrlaltca:
Too bad this has never been looked upon.
While trying to rebase this to master, i saw that commit 1c5bb3a91154008ecedec1830f96d7be35ef04ac definitely forces the `${ }` pattern, making this PR unimplementable anymore.

PR #817, opened 2016-06-15
Clone block variables

JPBetley:
Uses the same functionality from `cloneRow` to index variables inside blocks when using the `cloneBlock` function.

Before, variables inside cloned blocks would result in the same variable name used multiple times:

```
Template: cloneBlock('CLONE', 3)
${CLONE}
${variable}
${/CLONE}

Result:
${variable}
${variable}
${variable}
```

Now, in the same style of `cloneRow`, the variables are indexed:

```
Template: cloneBlock('CLONE', 3)
${CLONE}
${variable}
${/CLONE}

Result:
${variable#1}
${variable#2}
${variable#3}
```


troosan:
duplicate of #459

PR #830, opened 2016-07-05
Extends html parser

Slavenin:
Extends html parser


troosan:
I think most of those changes have been addressed. Feel free to reopen if more is needed.

PR #839, opened 2016-07-19
PHP7 issue with the src/PhpWord/Shared/String.php file

mlinuxgada:
Because in php7 you cant use the String class / in php5.6 we can/ I applied a fix for being able to use the component.
Its basically a cosmetic refactoring - renaming the needed namespaces used, the class name from String to smth like StringComponent .. and editing all the places its methods are called.


lukepolo:
Should be closed, already fixed in 13.

PR #863, opened 2016-08-16
Implement PageBreak for odt writer

cookiekiller:
Implemented page break for odt writer, works great.


cookiekiller:
I use it as it is, I added it using composer to my project extending phpword.

cookiekiller:
@Progi1984 I proposed and tested this. This patch is in use for some time now! I'd suggest to merge it.

Progi1984:
@cookiekiller Could you add some unit tests ?

troosan:
@cookiekiller I could not update your branch (readonly) so I created a new one (PR #1206)

PR #870, opened 2016-08-23
Padded the $args array to remove error.

reformed:
Line 101 was producing "Notice: Undefined offset:" errors despite the attempted error suppression. Padded the array to ensure these offsets are defined and removed the subsequently unnecessary suppression.


troosan:
also fixed in PR #1150 which is already merged

PR #875, opened 2016-08-29
hasDifferentPage() needs to check for first page footers

aoloe:


aoloe:
... otherwise first page footers are simply ignored... (if you did not also set a first page header...)

aoloe:
continuous integration is failing... but -- since the next patch submitted by somebody else also fails -- i guess it's because something is currently wrong in the code...

troosan:
merged PR #1116 already which does the same

PR #892, opened 2016-09-15
Add Style import

Benpxpx:
Add the Style import to MsDoc.php so Style is correctly referenced later on in the [switch statement](https://github.com/PHPOffice/PHPWord/blob/master/src/PhpWord/Reader/MsDoc.php#L1673) located here.


Progi1984:
Ever fixed in develop branch

PR #912, opened 2016-10-17
Fix namespace for Font in MsDoc

alsofronie:
This fixes the wrong namespace used for Font in MsDoc reader.


Progi1984:
Ever fixed in develop branch

PR #920, opened 2016-10-24
use missing namespace for PhpOffice\PhpWord\Style

mario-rivera:
the PhpOffice\PhpWord\Style\Font class wasn't found inside the namespace of the file

simply added use PhpOffice\PhpWord\Style to fix the issue


Progi1984:
Ever fixed in develop branch

Issue #927, opened 2016-10-29
Illegal string offset warning in PHP 7


likeuntomurphy:
Saving a document triggers an `Illegal string offset 'w:compatSetting'` warning, because `$this->settings['w:compat']` is initialized as an empty string instead of an empty array in `src/PhpWord/Writer/Word2007/Part/Settings.php`. This appears to be the case with PHP7 — PHP5 does not raise the warning.

germansokolov13:
I've just downloaded PHPWord 0.13.0 via Composer and it's still an empty string instead of an empty array there. Not merged into a release?

philwu:
I've just downloaded PHPWord 0.13.0 via Composer still get the problem.
Win10 build 14393+IIS
PHP 7.1.1

thibaultvanc:
FOR PHP 7.1

`
try {
      $objWriter->save($path);

    } catch (Exception $e) {
      // dd(123);
      // dd('cat');
    }
    return redirect('/'.$path);
`
WORK FOR ME

shad0wfir3:
@thibaultvanc can you expand on where you implemented this? currently getting the same error on php 7.1.2

sazal836:
Doc file created successfully 
but why this message is showing....

Warning: Illegal string offset 'w:compatSetting' in C:\xampp\htdocs\greek\vendor\phpoffice\phpword\src\PhpWord\Writer\Word2007\Part\Settings.php on line 173

Notice: Array to string conversion in C:\xampp\htdocs\greek\vendor\phpoffice\phpword\src\PhpWord\Writer\Word2007\Part\Settings.php on line 173

Warning: Invalid argument supplied for foreach() in C:\xampp\htdocs\greek\vendor\phpoffice\phpword\src\PhpWord\Writer\Word2007\Part\Settings.php on line 80

wqj97:
Still have problem today

Jaspur:
Same issue here. PHP 7.1.3

likeuntomurphy:
This has been merged into the `develop` branch and is slated for inclusion in [v0.14.0](https://github.com/PHPOffice/PHPWord/milestone/17). However, the [release is on hold indefinitely](https://github.com/PHPOffice/PHPWord/issues/848) because the project does not have an active maintainer.

[...]

PR #928, opened 2016-10-29
Fixes PHPOffice/PHPWord#927

likeuntomurphy:
The test works because the warning is only raised once `$phpWord->getCompatibility()->getOoxmlVersion()` is no longer null.


cookiekiller:
That's a good catch. I reviewed this change. @Progi1984

PR #931, opened 2016-11-02
+ add span tag for text style;

Mon1k:
text tag for style for text node


cookiekiller:
@Mon1k could you please have a look at the tests?

troosan:
Parsing of span should be OK now.
Please create a separate PR for the HTML writer table style

Issue #921, opened 2016-10-24
addImage base64 support?


koosvanderkolk:
In my code I have the image as a base64 encoded string. Of course I can write it to a temp file and then add it to my Word-file with addImage, but wouldnt it be more efficient if I could immediately feed addImage with my string? 

I did not really dive into the code, but the existence of the getImageStringData function seems to suggest that base64 encoding is already been used internally (http://phpoffice.github.io/PHPWord/docs/master/classes/PhpOffice.PhpWord.Element.Image.html).

troosan:
resolved in develop branch

koosvanderkolk:
Thanks!

nicdnepr:
Why in develop? Will it go to release?

nicdnepr:
Also its not work, I try dev-master and dev-develop and get error 
`Invalid image: data:image/png;base64` in `PhpWord/Element/Image.php` 385

padre:
@nicdnepr I have the same problem (I've also tried dev-master and dev-develop). Have you been able to fix it?

nicdnepr:
@padre No
I used url to return image

padre:
@nicdnepr ok, thanks.

Theoretically it should work, but after trying for hours there's no way :-(

@troosan, have you ever managed to use an image as a base64 encoded string?

markmorris:
@troosan Can confirm this dont work at all. Tried base64_decode as you suggested too and still nothing. Also tested my string online and its perfect and produces the image. Is there anyway we can add support for base64 encoded image?

either way i get a fatal error.

npth:
It works, if crop "data:image/png;base64," string from base64 code string, ex:
`$find = strpos($file_contents, ';base64,'); 
if ($find !== false) {
                $encoded_string = substr($file_contents, $find + 8);
            }
$section->addImage(base64_decode($encoded_string));`

PR #936, opened 2016-11-08
support for Image creation from string image data

troosan:
This change allows us to create an PhpOffice\PhpWord\Element\Image from a string representing the image data.
```php
        $source = file_get_contents(__DIR__ . '/../_files/images/earth.jpg');
        $image = new Image($source);
```
Of course in real life `$source` would come from a DB for instance.

Thanks to this you do not have to write the image data to file before passing it to the constructor.
This solves issue #921

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@ Maintenance release. This release is focused primarily on `TemplateProcessor`.
 
 
 
-0.12.0 (3 January 2015)
+v0.12.0 (3 January 2015)
 -----------------------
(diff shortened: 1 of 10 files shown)


troosan:
@myrecs can you post a sample code that reproduces your problem?

myrecs:
there is nothing special
`
$source = file_get_contents(*here full path to jpg file*);
$section->addImage($source, array('height'=> 400,'alignment' => 'center'));`

troosan:
I added the following in any sample (from the samples directory) and it works.
```php
$source = file_get_contents('resources/_earth.jpg');
$section->addImage($source, array('height'=> 400,'alignment' => 'center'));
```
I'm using phpword 0.14 and opening the word file with Word 2016

PR #937, opened 2016-11-08
add support for Image creation from string image data

troosan:
This change allows us to create an PhpOffice\PhpWord\Element\Image from a string representing the image data.
```php
        $source = file_get_contents(__DIR__ . '/../_files/images/earth.jpg');
        $image = new Image($source);
```
Of course in real life `$source` would come from a DB for instance.

Thanks to this you do not have to write the image data to file before passing it to the constructor.
This solves issue #921

### DIFF
--- a/docs/elements.rst
+++ b/docs/elements.rst
@@ -232,7 +232,7 @@ To add an image, use the ``addImage`` method to sections, headers, footers, text
 
     $section->addImage($src, [$style]);
 
-- ``$src``. String path to a local image or URL of a remote image.
+- ``$src``. String path to a local image, URL of a remote image or the image data, as a string.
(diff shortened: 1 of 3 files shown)


Progi1984:
@troosan Good job in the MR. Tests units & Code are 👍 . Could you add some documentation, before the merge ? Thanks

troosan:
@Progi1984 I updated the documentation but for whatever reason the build is failing now. The build is taking forever and finally php-doc fails to be executed.
Is there something wrong with what I committed or is there something wrong with Travis?

Progi1984:
@troosan Perfect. PhpDoc has some bugs currently, I don't care about it in my review.

PR #940, opened 2016-11-10
Update to latest version of PclZip lib

nickknissen:
This fixes the error `PHP 4 constructors are now deprecated` generated
when running tests with PHP7


troosan:
fixed in other PR

PR #941, opened 2016-11-10
Update to latest version of mpdf in test requirements

nickknissen:
This fixes the PHP 4 constructor error when running the tests with PHP7

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "phploc/phploc": "2.*",
         "dompdf/dompdf":"0.6.*",
         "tecnickcom/tcpdf": "6.*",
-        "mpdf/mpdf": "5.*"
+        "mpdf/mpdf": "6.*"
     },
     "suggest": {
         "ext-zip": "Allows writing OOXML and ODF",


cookiekiller:
This will raise the current PHP min requirements to 5.4.0. Current minimum is 5.3.3.
I'm not sure about dropping the 5.3.3 support yet.

@Progi1984 what's your opinion about it? I'm would suggest updating some dependencies as well as Zend to disfavor of old PHP support. I'd suggest a poll to determine the communitiy's opinion.

Progi1984:
@cookiekiller For me, changing the PHP version is [a major change](http://semver.org/#spec-item-8) so that must be done with care.

nickknissen:
PHP4 constructor are "only" deprecated and can still be used in PHP7. So this is not necessary to be merged in right now.

marcomenzel:
I'm not sure if that matters, but the requirement of ZF2.4.* is at least PHP >=5.3.23

troosan:
fixed in 0.14, I put a choice between 5.* || 6.*
Depending on the PHP version you run composer update with, the correct version will be taken.

PR #960, opened 2016-12-18
Add style ID to font style, so indesign import works

DaanMeijer:


troosan:
Is this still needed? In the develop branch a styleId is now being set if $paragraphStyle is not null.
Maybe we can keep the addition of w:customStyle, although only in the else part, otherwise it breaks the titles.

Issue #622, opened 2015-09-15
Reader/writer for Comments and Track-changes document parts


rvanlaak:
The document graph also contains comments and track changes. We would like to be able to also read those, so we can deduct statistics from it:

https://msdn.microsoft.com/en-us/library/bb266220(v=office.12).aspx

Can't find much info about why it isn't there, no issues related to comments or track changes can be found. If you guys tell me where to start then I'll try to make a PR for the `Reader` part.

chris-fa:
Hello, did you manage tu turn on track changes in a document from PHPWord?

rvanlaak:
Yes, you can do that by editing `word/settings.xml`. The track changes is related to `w:proofState`. 

Other interesting settings are `w:spelling` and `w:grammar`, which you can set to `dirty` to for instance re-enable all spelling checks.

chris-fa:
@rvanlaak thanks for your fast response, really appreciate it :)

Can you point me to any type of documentation or resource regarding this xml file? I'm not sure if it is related to the OOXML standard or something else (I'm kind of new to editing word documents from PHP 😅 )?

rvanlaak:
The OOXML specs are documented quite decently, maybe this gives you some info already: http://officeopenxml.com/anatomyofOOXML.php

Otherwise try MSDN https://msdn.microsoft.com/en-us/library/documentformat.openxml.wordprocessing.proofstate(v=office.14).aspx

chris-fa:
@rvanlaak Thanks for the links. Are you sure the "Track Changes" is related to `w:proofState`? I can only find info on grammar and spell checking on every site :(

rvanlaak:
I think it's `w:trackRevisions`: http://openxmldeveloper.org/blog/b/openxmldeveloper/archive/2007/01/19/features-of--track-changes-revisions-explored.aspx

chris-fa:
Thanks, I found this blog post in case anybody needs more info: http://openxmldeveloper.org/discussions/formats/f/13/p/800/2032.aspx#2032

chris-fa:
Ok so, since there is  no reader/writer for tracking changes, I'll try making one and submit a PR

chris-fa:
@rvanlaak Hello, I just submitted PR #966 that solves this. It has about 3 TODOs regarding documentation, which I need some help, if possible.

Also, I don't know the maintainers, so if you could mention them, that'd be nice 😄

troosan:
[...]

PR #966, opened 2016-12-20
Add metadata for w:trackRevisions setting

chris-fa:
### Description
Adds a metadata class `Revisions` that let's the user enable `w:trackRevisions` on the `word/settings.xml` file

### Reason
The settings writer didn't have a way to enable the setting for tracking revisions

### Impact
Users can now enable track revisions option

### Issue
Issue #622 

### Checklist
- [x] Did I write tests for this change?
- [ ] Is documentation complete?

### DIFF
--- a/src/PhpWord/Metadata/Revisions.php
+++ b/src/PhpWord/Metadata/Revisions.php
@@ -1,0 +1,58 @@
+<?php
+/**
+ * This file is part of PHPWord - A pure PHP library for reading and writing
+ * word processing documents.
+ *
+ * PHPWord is free software distributed under the terms of the GNU Lesser
+ * General Public License version 3 as published by the Free Software Foundation.
+ *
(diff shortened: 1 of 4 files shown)


rvanlaak:
Nice! Our use case was to actually write the settings, so did something like this:

```php
public function setProofState(\SimpleXMLElement $settings)
{
    $settings = $this->unsetByXpath($settings, 'w:proofState');

    $proofState = $settings->addChild('w:proofState');
    $proofState->addAttribute('w:spelling', 'dirty', self::NS);
    $proofState->addAttribute('w:grammar', 'dirty', self::NS);

    return $settings;
}
```

Unfortunately I think making a writer is outside of the scope of this project (ping @Progi1984 ).

troosan:
please check PR #1098 for new impl

PR #903, opened 2016-09-28
Composer update

jarrettj:
Use the latest zend framework 2.\* modules. Fixes #865.

PR #973, opened 2017-01-14
loosen zendframework constraints, extend building

marcomenzel:
As mentioned here https://github.com/PHPOffice/PHPWord/pull/903#issuecomment-272100136 I would like to loosen the zendframework constraints

While depending on `zendframework/* ^2.4` ...
* It would not allow a downgrade to ZF >= 2.0 and <2.4
* PHPWord would stay PHP 5.3 compatible while using ZF 2.4
* You can use ZF 2.5 with "newer" PHP versions

While doing that I would like to extend the build on travis with php5.3 and --prefer-lowest option on composer install to test that the oldest possible composition of components will still work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: php
 
 php:
     - 5.3
+      env:
+        - PREFER_LOWEST=true
     - 5.4
     - 5.5
     - 5.6
@@ -25,7 +27,8 @@ before_install:
 before_script:
     ## Composer
     - composer self-update
-    - composer install --prefer-source
+    - if [ -z "$PREFER_LOWEST" ]; then composer install --prefer-source; fi
+    - if [ "$PREFER_LOWEST" == 'true' ]; then composer install --prefer-source --prefer-lowest; fi
     ## PHPDocumentor
     - mkdir -p build/docs
     - mkdir -p build/coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
     ## Composer
     - composer self-update
     - if [ -z "$PREFER_LOWEST" ]; then composer install --prefer-source; fi
-    - if [ "$PREFER_LOWEST" == 'true' ]; then composer install --prefer-source --prefer-lowest; fi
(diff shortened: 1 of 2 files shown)


marcomenzel:
Mh. There are some (most) tests failing because of
* bugs in lowest possible version of phpunit/phpunit 3.7.0
* bugs in lowest possible version of phpoffice/common 0.2.0

I'm not sure how to handle this :-(

maglnet:
I'd suggest: remove the build system adjustment and keep the corrected dependency version constraints.

marcomenzel:
@troosan 
do you have any plans to loosen these zendframework constraints (e.g. `zendframework/* ^2.4`) or do you have any hints what can be done to help this project achieve this?

(My personal objective is to get rid of my ugly fork of PHPWord which uses just the newer versions of the zendframework packages.)

troosan:
@marcomenzel At some point we'll have to to be compatible with PHP 7.1
Currently a couple of dependencies are generating a lot of warnings and making the automated build fail for PHP >= 7.0 (it's just warnings but still).

Currently PHPWord requires PHP 5.3.3, so ideally we should use zend 2.2.* max, like this
image
But instead we have
image

As of Zend 2.3.x requires PHP 5.3.23 (or I change PHPWord to require 5.3.23 instead)

To correctly support PHP 7.1 we'll have to drop official support for PHP 5.3
I guess I will do this when updating to version 1.0 as this is really a major change for people still running PHP 5.3 (I know it's so old already!)

I had another discussion about this here #1026

marcomenzel:
Thanks for these detailed explanation, @troosan!

So, at all ~0.13 will stay at the php 5.3.3 requirement and a coming 1.0 release will require (?) php 7.1.
Playing with these thoughts means that you would reommend to use a personal fork if I want/have to use php 5.6 and zendframework > 2.4?

I do not remember my findings on the breaking builds, but I think it was combined of something like _phpoffice/common is requiring php >= 5.3.0 instead of requiring ^5.3.0_ and some other breaking dependencies.

jarrettj:
Thanks for the progress on this issue. My two cents (like it matters:) would be to move forward and shoot PHP 5 in the face.  :-D

troosan:
fixed with #1138

Issue #976, opened 2017-01-21
Wrong paper size computation


troosan:
Concerning PHPWord/Style/Paper, the conversion from mm -> twip is incorrect.
At line 163, the rate is set to 56.5217 whereas it should be 56.692913386.

It seems that for A4, the end result is correct (thanks to rounding I suppose) but if I add an entry for B5 in the list of sizes, the computed value is 1 mm off once you open it in word, and as a consequence word is not detecting it as B5 but as a custom size.

Solutions are to either

- just correct the rate
- use the PHPWord/Shared/Converter which does the job correctly
- manually set the paper size, not using the official names (A4, B5, ...), but I haven't found how ...

Shall I create a pull request for this or is there another way to use custom sizes?

PR #977, opened 2017-01-21
fix paper size and add tests for Paper class

troosan:
fix for issue #976 
I opted for the call to the Convertor as I'm not for duplicating code :-)

### DIFF
--- a/src/PhpWord/Style/Paper.php
+++ b/src/PhpWord/Style/Paper.php
@@ -17,6 +17,8 @@
 
 namespace PhpOffice\PhpWord\Style;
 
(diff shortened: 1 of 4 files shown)


troosan:
So, I broke the tests ...
It seems to me the constants defined in Section are wrong too.
DEFAULT_WIDTH is set to 11870 for instance

Using http://www.unitconversion.org/typography/twips-to-millimeters-conversion.html for instance, when you convert 11870 in mm, you end up with 209.373611111, which should in fact be 210 (A4 format is 210 x 297)

As far as I can see those values were changed by @ivanlanin. I guess all this started from the wrong conversion rate used in the Paper class.

PR #982, opened 2017-01-25
Allow to change cell width unit

guillaume-ro-fr:
The width default cell unit was `dxa`.
With this patch, the width unit of a cell can be changed in the cell style.


guillaume-ro-fr:
For the failed tests results, see #983

guillaume-ro-fr:
Error of branch, see #986

PR #986, opened 2017-01-28
Allow to change cell width unit

guillaume-ro-fr:
The width default cell unit was `dxa`.
With this patch, the width unit of a cell can be changed in the cell style.


troosan:
Thanks, very similar fix committed on the development branch

Issue #983, opened 2017-01-26
getimagesize SSL Problem


guillaume-ro-fr:
The ImageTest::testUnsupportedImage() PHPUnit failed because of the certificate error with PHP on samples.libav.org.

```
php >getimagesize('http://samples.libav.org/image-samples/RACECAR.BMP');
PHP Warning:  getimagesize(): SSL operation failed with code 1. OpenSSL Error messages:
error:14090086:SSL routines:ssl3_get_server_certificate:certificate verify failed in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. getimagesize() php shell code:1

Warning: getimagesize(): SSL operation failed with code 1. OpenSSL Error messages:
error:14090086:SSL routines:ssl3_get_server_certificate:certificate verify failed in php shell code on line 1

Call Stack:
    6.6319     354064   1. {main}() php shell code:0
    6.6319     354064   2. getimagesize() php shell code:1

PHP Warning:  getimagesize(): Failed to enable crypto in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. getimagesize() php shell code:1

Warning: getimagesize(): Failed to enable crypto in php shell code on line 1

Call Stack:
    6.6319     354064   1. {main}() php shell code:0
    6.6319     354064   2. getimagesize() php shell code:1

PHP Warning:  getimagesize(http://samples.libav.org/image-samples/RACECAR.BMP): failed to open stream: operation failed in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. getimagesize() php shell code:1

Warning: getimagesize(http://samples.libav.org/image-samples/RACECAR.BMP): failed to open stream: operation failed in php shell code on line 1

Call Stack:
    6.6319     354064   1. {main}() php shell code:0
    6.6319     354064   2. getimagesize() php shell code:1
```

troosan:
I have a fix ready for this.
The change I did is to change the way the image is loaded when the protocol is https.
In that case I load the image with file_get_contents, with which it is possible to pass the correct certificate using stream_context_set_default(...). It seems this is not possible when using getimagesize(...)

guillaume-ro-fr:
Fixed with 988

PR #987, opened 2017-01-29
Fix image loading over https

troosan:
this fixes issue #983 
As images are now loaded using file_get_contents, it is possible to modify the stream context using stream_context_set_default(...).
In ImageTest I'm using 
```php
        $arrContextOptions = array(
            "ssl" => array(
                "verify_peer" => false,
                "verify_peer_name" => false,
            ),
        );
        stream_context_set_default($arrContextOptions);
```
which you should obviously not do in a production environment, instead you should pass the certificate to use to connect to the remote url.


troosan:
hum ... wrong branch!

PR #993, opened 2017-02-02
composer: bump to PHP 7.0, only actively developer version

TomasVotruba:
See: http://php.net/supported-versions.php

image


TomasVotruba:
Ping

PR #1013, opened 2017-03-02
Work with PHPWord

tanushgupta:
Initial pull request


troosan:
The same change was done in PR #821

PR #1017, opened 2017-03-09
Introduce deleteRow() method for TemplateProcessor

simivar:
Introduce `deleteRow()` method for TemplateProcessor method that works fine with cell's spanning multiple rows.


FBnil:
Checked the code and it is superseeded by https://github.com/PHPOffice/PHPWord/pull/1147
Which also contains `deleteRow()` and a test.

@simivar We are rivals in code! :) 

Your test document does not contain a cell containing multiple rows (double as high as the previous one containing the `<w:vMerge w:val="restart"/>`).
```
+--------+--------+
| ${cell1} | ${cell2} |
+           +--------+
|              | ${cell3} |
+--------+--------+
```
where you select to delete `${cell1}` and  `${cell2}` and `${cell3}` should be gone too, but are not. 
I suspect that the `CloneRow()` never worked as intended. I am fixing it in my branch.

FBnil:
Ok, in LibreOffice (5.4.1 and 5.4.2)
The Rowspan is `<w:vMerge w:val="continue"/>`, and does not match `cloneRow()`'s  `('#<w:vMerge w:val="continue" />#', $tmpXmlRow)`
So the easy fix is adding a question for the space:
`('#<w:vMerge w:val="continue" ?/>#', $tmpXmlRow)  // <w:vMerge w:val="continue"/>`

FBnil:
I also found another bug:
```
+---+---+---+
| A |   |   |
|   |   +---+
|   | B |   |
|   +---+---+
| C |   |   |
+---+---+---+
```
When cloning `A`, the End-of-span marker is found at point `B`, not at point `C`. So either a counter should be used, or we assume the user is smart enough to give an extra $span parameter in the cloneBlock().

What do you think the solution should be?

Some nice sideeffects that always were in the code: `cloneRow(B)` does not clone block A, although there seems to be a hint that it tried.

manukieli:
I the code working with the version 0.16.0
It's a very good idea.

Issue #546, opened 2015-06-07
Create password protected docx file


kasinau:
Hi, I want to create a docx file from a template and protect it to be able to open only entering a password.
Is it possible with PHPWord or no?
Thank you!

DsXack:
:+1: 
It no possible now(

https://github.com/PHPOffice/PHPWord/blob/3f136501b3dfea94766a704e17782c8a942a7b9b/src/PhpWord/Metadata/Protection.php#L25

troosan:
no, that's not possible. you can only make a document (or parts of it only) read-only.
Forbidding to open a document is much more difficult than that as the whole document needs to be encrypted.

devjob:
I have not protected template Word document.
I need to edit it and then to set protection with password to avoid changes.
How could I do it?

PR #1019, opened 2017-03-13
enable password protection (Word)

mariahaubner:
This pull request will solve #546 

It enables PHPWord to create password protected docx-files which MS Word can open and unlock. It was tested on MS Word 2010.

To implement the password protection, we translated the C#-code from https://blogs.msdn.microsoft.com/vsod/2010/04/05/how-to-set-the-editing-restrictions-in-word-using-open-xml-sdk-2-0/ 

The actual password-hash generation is located in /Writer/Word2007/Part/Settings.php as it is Word-specific and cannot be used for the other writers. 
I hope the comments left in the code and the link to the C# source clarify what is happening. But if there are any questions or remarks, feel free to share them.

Cheers, Maria


crevillo:
hello, what's is the state of this? still valid pr? i tried to test with master branch but there are some conflicts.

src/PhpWord/Shared/Microsoft/PasswordEncoder.php
```diff
@@ -0,0 +1,205 @@
...
+ * @copyright   2010-2017 PHPWord contributors
+ * @license     http://www.gnu.org/licenses/lgpl.txt LGPL version 3
+ */
+
+namespace PhpOffice\PhpWord\Shared\Microsoft;
+
+/**
+ * Password encoder for microsoft office applications
+ */
+class PasswordEncoder
```
troosan: @maartenba @Progi1984 Would it make sense to move this class to the PHPOffice common project? I guess it could be reused in PhpSpreadsheet

maartenba: I think it would make sense to do, yes.

Progi1984: The same as @maartenba 

Poke @MarkBaker & @PowerKiki

troosan: @Progi1984 @maartenba @MarkBaker @PowerKiKi I created a PR in Common for this.

PowerKiKi: I am actually not quite sure how this should/could be used in PhpSpreadsheet...

FYI, the PR: https://github.com/PHPOffice/Common/pull/15

troosan: It seems indeed that pre-2013 excel sheets use a very simplistic password hashing which is completely different.
In more recent specs, it also uses something very similar to what is used in word. Not sure the algorithm is the same though.

https://www.thespreadsheetguru.com/blog/2014/8/20/understanding-excels-password-security-methodology

troosan: @PowerKiKi I see you have the pre-2013 hashing implemented already in PhpSpreadsheet/Shared/PasswordHasher, this would be a more modern password hashing mechanism

PR #1040, opened 2017-04-04
Add support for cellSpacing for tables

dox07:
changes for continuous integration


troosan:
@dox07 would you find some time to do the following so we can merge this feature in?
- [ ] write unit tests to cover this
- [ ] update the documentation

PR #1050, opened 2017-04-25
Allow reading/replacing mailmerge

dlwatersuk:


dlwatersuk:
Regex and functions taken from an earlier attempt at adding this I found in the pull requests, little bit of reworking and it looks to work in the use cases I've tested.

FBnil:
@dlwatersuk Very useful, however you should make a subclass, like:
`class MailMergeTemplateProcessor extends \PhpOffice\PhpWord\TemplateProcessor { ....`
and only put the functions there that you actually change/add, AND name that file `MailMergeTemplateProcessor.php`. This will make it easier to use both the old and the new class.

Also think broader, you can mailmerge, but actually you are using FIELDS.
http://www.officeopenxml.com/WPfields.php

 It also lacks a MailMergeTemplateProcessorTest.php (just look at TemplateProcessorTest.php). It is actually hard work to add that. For example, the specifications say you can have "One or more <w:instrText> elements, which, together, contain a complete field.", so you need a testcase to handle such a fragmented field.

You also do not handle `dirty`, where you unset that field because you just replaced it, so it is current information, right?
I also think you use way too much code. Cleanup setMailMergeValue() it does nothing at the moment. make a new function for the regexp that you use in many places. Get rid of setMergeData and just add it as a parameter to doMerge. Do not pass $replace[] around, you may understand it, but the code will have to be supported later on, and that could be someone else.
I also see only support for w:fldChar and not fldSimple.

Can you post some information about mailmerging?

eborned:
I've been testing this class for a new project. For the most part, it works. But for larger documents, the `parseMergeSection` function return `null`, which means the PREG library errored out. In my case it was the `PREG_JIT_STACKLIMIT_ERROR` error, which must be related to the backtracking that is used in the regex, but I'm unsure where to start.

Wouldn't it be better to just replace all the regex matches and replacements with DomDocument?

PR #1059, opened 2017-05-15
Support comments in docx documents

troosan:
This will allow you to add comments in your generated documents.
To try it out, run the Sample_37_Comments.php

This will generate the following

image


src/PhpWord/Element/AbstractElement.php
```diff
@@ -108,13 +108,27 @@
...
      * @var bool
      */
     protected $collectionRelation = false;
 
     /**
+     * The start position for the linked comment
+     * 
+     * @var Comment
+     */
+    protected $commentStart;
```
rvanlaak: Should this be part of `AbstractElement`? Seems it should be part of `Comment` to me. Or is it because comments should be attached to an element? If so, maybe the attribute name could be clarified just a bit more? 👍

troosan: @rvanlaak Indeed, the Comment starts at an element and ends at another one.

The node in the generated XML is named 'w:commentRangeStart and 'w:commentRangeEnd, maybe I could rename it to match this closer.

On the Comment itself, I'm keeping a reference back to the Element on which it starts/ends
 private $startElement;
 private $endElement;

troosan:
I messed up the merge on this branch. I'm deleting this branch and pushing to a new one.

hrydi:
i have trouble for using this feature (comment), i have couples of comments and want to attach it to single text. but always outputing one comment instead (i'm using loop for this).
can you help me ?
and sorry for my bad english :smile: .
this library is awesome btw, thank you.

troosan:
@hrydi I'm afraid I had not foreseen to add multiple comments on a same text element. Which is indeed wrong :-(
I'll see what I can do about that, but it could take some time!

hrydi:
@troosan thanks for the reply,
I will wait for the updated :smile: 
while waiting, I will test in my local of course hehehe

cambraca:
@troosan hey hey, has there been any update on multiple comments? (it's hard to google this, the word "comment" pops up everywhere on the web!)

Thanks!

PR #1068, opened 2017-05-29
Add possibility to control the footnote number

troosan:
This will allow you to use another numbering for the footnote reference.
- 1, 2, 3 (default)
- I, II, III
- i, ii, iii
- A, B, C
- ...

You can also decide to put all footnotes at 
- the end of the document
- the end of the section
- the end of the page (default)

You can also decide when to restart from 1
- continuous (default)
- each Section
- each Page

And finally you can decide not to start with 1

### DIFF
--- a/docs/elements.rst
+++ b/docs/elements.rst
@@ -333,10 +333,27 @@ On text:
     $footnote = $section->addFootnote();
     $footnote->addText('Footnote text.');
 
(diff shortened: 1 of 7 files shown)


troosan:
I should probably add the possibility to set the properties for the whole document, so you don't have to repeat it for each section ...

PR #1070, opened 2017-06-02
Whether use CDATA to handle XML

lampnick:
When we use setValue() function,if we pass the $replace parameter with special character such as &,<, etc.The generated word document can't open with the error message like "illegal character" in /word/document.xml,line：2，column：925564.So,I suggest to use default CDATA to surround the $replace  param.


FBnil:
@lampnick 
How does MSWord/LibreOffice handle cdata? Is it still editable (wordbreaks, paragraphs) and does the spellchecker still work?
How about a subclass that overwrites `protected static function ensureUtf8Encoded($subject)` to do cdata instead of UTF8 encoding? What is wrong with the current `Settings::isOutputEscapingEnabled()` implementation?

lampnick:
@FBnil It's passed by php5.3,5.4,5.5, but  in php5.6  throwed the following exception:
"1) PhpOffice\PhpWord\Element\ImageTest::testUnsupportedImage
Failed asserting that exception of type "PhpOffice\PhpWord\Exception\InvalidImageException" matches expected exception "\PhpOffice\PhpWord\Exception\UnsupportedImageTypeException". Message was: "Invalid image: http://samples.libav.org/image-samples/RACECAR.BMP"."

FBnil:
@lampnick in ./tests/PhpWord/Element/ImageTest.php you have this function:
```
    /**
     * Test unsupported image
     *
     * @expectedException \PhpOffice\PhpWord\Exception\UnsupportedImageTypeException
     */
    public function testUnsupportedImage()
    {
        //disable ssl verification, never do this in real application, you should pass the certiciate instead!!!
        $arrContextOptions = array(
            "ssl" => array(
                "verify_peer" => false,
                "verify_peer_name" => false,
            ),
        );
        stream_context_set_default($arrContextOptions);
        $object = new Image('https://samples.libav.org/image-samples/RACECAR.BMP');
        $object->getSource();
    }
```

See that `@expectedException` in the comments? It is interpreted as "this function will throw an exception of this type" during the testing.

Now, in ./src/PhpWord/Element/Image.php we have two nested "trow" commands in __construct()  (which is what you call when you do a new Image()  ) and checkImage() has the 
```
        if (!is_array($imageData)) {
            throw new InvalidImageException(sprintf('Invalid image: %s', $this->source));
        }
```
Which could happen, when, for example, the website samples.libav.org is too busy and does NOT give you an image. So, During NOT USA daytime hours, edit "certiciate " into "certificate" and push it again (in the hope, this time, it goes all ok).

lampnick:
@FBnil Knowbie, Thank you very much!

FBnil:
I DO see one 'but', and that is the following:

First we do a 
```
 $replace = '<![CDATA[' . $replace . ']]>';
```
And then we do a 
```
$replace = $xmlEscaper->escape($replace);
```

This basically means:

(The escape() function does this: `return htmlspecialchars($input, ENT_QUOTES, 'UTF-8');`)

`'<![CDATA[' . $replace . ']]>';` becomes `'&gt;![CDATA[' . $replace . ']]&lt;';` BEFORE the insertion into the document.

You solve this by adding the CDATA wrapping after the escape() function.

What you also need is a TestCase, 
I added [reflections in my testcases](https://github.com/FBnil/PHPWord/blob/89d60bb082897a38c10dbbe84279ee9e6b69149b/tests/PhpWord/TemplateProcessorTest.php) (see the peek and poke functions), this allows me to rewrite the private tempDocumentMainPart variable that holds the document xml:
`$this->poke($templateProcessor, 'tempDocumentMainPart', $newXmlStr);`
And read it out with peek:
`$this->peek($templateProcessor, 'tempDocumentMainPart')`
so you can make many testcases and combinations without creating yet another word document.

lampnick:
@FBnil Thank you! I'll try.

troosan:
All you need to do is call
```php
Settings::setOutputEscapingEnabled(true);
```

PR #1116, opened 2017-08-02
Fix: different footer for first page

ctrlaltca:
Sample 12 shows how to add a different header for the first page:
```
// Add first page header
$header = $section->addHeader();
$header->firstPage();
...
// Add header for all other pages
$subsequent = $section->addHeader();
```
This is supposed to work also for footers (since `Header` itself is a subclass of `Footer`), but it's currently broken because `Section::hasDifferentFirstPage()` only checks if a first-page specific header exists, but doesn't check for a first-page specific footer, too.

This PR adds the missing check, fixing the rendering of footers with type `Footer::FIRST`.

### DIFF
--- a/src/PhpWord/Element/Section.php
+++ b/src/PhpWord/Element/Section.php
@@ -181,6 +181,11 @@ class Section extends AbstractContainer
                 return true;
             }
         }
+        foreach ($this->footers as $footer) {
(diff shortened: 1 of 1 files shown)


ctrlaltca:
As a workaround until this gets fixed, creating an empty header for the first page is enough to get the footers rendered correctly.

Issue #1044, opened 2017-04-12
Zend framework 3 and phpword issue


haolle:
Just downloaded Zend framework 3 skeleton applications  by "composer", and tried to add phpword component. Unfortunately, the skeleton application comes with Zend-escaper 2.5.2, while phpword requires 2.4.*. Couldn't get phpword installed. Any suggestions? Thanks.

jarrettj:
I'm also waiting :|

Yamakasi:
Here it fails on Zend-validator instead of escaper for 2.4.x

troosan:
fixed with #1138

Yamakasi:
If you ask me this is not fixed yet:

`# composer require phpoffice/phpword
Do not run Composer as root/super user! See https://getcomposer.org/root for details
Using version ^0.13.0 for phpoffice/phpword
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for phpoffice/phpword ^0.13.0 -> satisfiable by phpoffice/phpword[v0.13.0].
    - phpoffice/phpword v0.13.0 requires zendframework/zend-validator 2.4.* -> satisfiable by zendframework/zend-validator[2.4.0, 2.4.0rc1, 2.4.0rc2, 2.4.0rc3, 2.4.0rc4, 2.4.0rc5, 2.4.0rc6, 2.4.0rc7, 2.4.1, 2.4.10, 2.4.11, 2.4.12, 2.4.13, 2.4.2, 2.4.3, 2.4.4, 2.4.5, 2.4.6, 2.4.7, 2.4.8, 2.4.9] but these conflict with your requirements or minimum-stability.


Installation failed, reverting ./composer.json to its original content.
`

Yamakasi:
Anyone able too look into this ?

troosan:
That's because it is fixed in the current development branch. It will be in 0.14

Yamakasi:
OK, thanks for that!

Can I update using composer to the development branch ?

PR #1122, opened 2017-08-16
Fix requirements from zend to be looser.

chasenyc:
Resolves issue #1044:
Switch  dependencies from Asterisk version numbers to [Tilde Version Range.](https://getcomposer.org/doc/articles/versions.md#tilde-version-range-)


troosan:
Fixed with #1138

PR #1125, opened 2017-08-21
Fix Some Problem When Using DomPDF

MikeMaldini:
When I Using DomPDF to Convert Word , I Find Some Problems So I Try To Fix It


troosan:
merged in develop branch

PR #1137, opened 2017-09-08
[Doc] elements add reference to constant

brutalcrozt:
referencing constant (PHPWord_Style_ListItem) to the source on github via permalink.

Please let me know if the sphinx syntax is wrong, this is my first time using it :smile:


troosan:
I'm not sure the idea was to have a link to the code. I guess the idea was to just print the Class name. So it should be PHPWord\\Style\\ListItem
Anyway, the link does not seem to work when it contains slashes or _

brutalcrozt:
@troosan i got it, but my idea is just to simplify when reading the doc, if the reader want to know the available constant.

brutalcrozt:
maybe any idea to listing available constant on the documentation?

PR #1142, opened 2017-09-18
Allow zendframework/zend-escaper up to 2.5.x ?

Artusamak:
Hello, thanks for this libray, i'm eager to be able to test it against a Drupal 8 project.
In order to do that, we need to be able to meet a minimum version requirement.
Drupal requires zendframork/zend-feeds that requires zendframework/zend-escaper to be in ^2.5.
Your actual version constraint is against 2.4.* and of course 2.4.x and 2.5.x and concurrent releases.
I'm starting this thread to see if the 2.5.x branch is working and hope it may be integrated in your core release soon.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "require": {
         "php": ">=5.3.3",
         "ext-xml": "*",
-        "zendframework/zend-escaper": "2.4.*",
+        "zendframework/zend-escaper": "2.5.*",
         "zendframework/zend-stdlib": "2.4.*",
         "zendframework/zend-validator": "2.4.*",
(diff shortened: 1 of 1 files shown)


troosan:
@Artusamak did you try the develop stream? I did this change already, changed it to ^2.2 actually.
So depending on the php version you run composer with it should take the latest compatible zend library version.

Artusamak:
Yes, switching to the develop branch did the trick, thank you!
For those trying to attempt the same goal, beware that you need to install PHPWord first to have zendframework/zend-stdlib (2.7.7) installed (and not the 3.x version used by Drupal).

PR #1143, opened 2017-09-22
Doc elements, fix syntax code-block::php

brutalcrozt:
Use the correct syntax for code-block, remove the comment in code-block.
currently render like so
phpword

### DIFF
--- a/docs/elements.rst
+++ b/docs/elements.rst
@@ -138,9 +138,9 @@ Page breaks
 There are two ways to insert a page breaks, using the ``addPageBreak``
 method or using the ``pageBreakBefore`` style of paragraph.
 
-:: code-block:: php
+.. code-block:: php
 
-    \\$section->addPageBreak();
+    $section->addPageBreak();
 
 Lists
 -----


brutalcrozt:
can comeone merge this?

FBnil:
@brutalcrozt There is no maintainer for PHPWord at the moment. Yes, there are two people that help out but nobody wants to take on the daunting task to evaluate and merge all changes to produce 0.14.0  You are welcome to apply to that position, and merge your own code in the process.

brutalcrozt:
@FBnil Thanks for the explanation, it my build was fail because fail to install php `5.3`. I think it will be wise if waiting your PR merged, Then i can retrigger the build process.

Looks like your PR is more urgent than mine, thank for your work. I also happy if i could be the maintainer, But if i am alone, it think i will break thing rather than fix a thing. But if with you that will be different story. I already subscribe your PR, waiting it be merged.

troosan:
applied manually in other branch

PR #1145, opened 2017-09-26
Working Block functions

FBnil:
I removed the regexp (although I had a working regexp, it seems PHP7 does not like multiple non-greedy patterns. The need to anchor the query to <?xml> is very dirty. Strange, because http://regex101.com/ and Perl have no problem with the same regexp. I'm quite good at regular expressions, so it pains me not using them). 
So I implemented the findBlockStart() and findBlockEnd() (same way that cloneRow() is implemented) that searches upto a paragraph change <w:p>, This has been tested with LibreOffice generated docx (that features <w:p>) and real MSOffice documents (that features extra parameters, nb: <w:p w:rsidR="00AC46F7" w:rsidRDefault="00AC46F7" w:rsidP="00AC46F7">). As well as a few new testcases to cover the new functionality.

What is new?

    cloneBlocks with variables inside now expand like cloneRows, with #n at the end (and you can get the old behavior back with an extra parameter)
    Block functions now use the same method as Row functions to detect start/end (marginally faster)
    Block functions now return sensible information instead of void
    you can throw exceptions if you can not find a block (enabled by an extra function parameter)
    getBlock() implemented
    Complete coverage of unit tests (including obvious missing tests) for Block functions. the good stuff.
    Multiple same named blocks behave as before (you can replace them one by one)
    Went through phpcs and phpunit to test everything. All tests pass* All formatting passes*

Files modified:
./src/PhpWord/TemplateProcessor.php
./tests/PhpWord/TemplateProcessorTest.php

OpenTemplateProcessor is a subclass that allows access to the private fields in TemplateProcessor. This is required for good testing of the output results (read the generated internal xml).

Tested with:
phpunit --bootstrap autoload.php phpoffice/PHPWord/tests/PhpWord/TemplateProcessorTest.php
phpcs ./src/PhpWord/TemplateProcessor.php ./tests/PhpWord/TemplateProcessorTest.php --standard=PSR2  

Todo: Tested under PHP 7.0.22 maybe try older versions too (although no PHP7-only code has been used). Also: new here, so if I'm doing things wrong, just tell me.

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -321,60 +321,101 @@ class TemplateProcessor
      * @param string $blockname
      * @param integer $clones
      * @param boolean $replace
+     * @param boolean $incrementVariables
+     * @param boolean $throwexception
      *
      * @return string|null
      */
-    public function cloneBlock($blockname, $clones = 1, $replace = true)
+    public function cloneBlock($blockname, $clones = 1, $replace = true, $incrementVariables = true, $throwexception = false)
     {
-        $xmlBlock = null;
-        preg_match(
-            '/(<\?xml.*)(<w:p.*>\${' . $blockname . '}<\/w:.*?p>)(.*)(<w:p.*\${\/' . $blockname . '}<\/w:.*?p>)/is',
-            $this->tempDocumentMainPart,
-            $matches
-        );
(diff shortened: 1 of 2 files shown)


FBnil:
fixing Scrutinizer issues, will offer a pull after fixes

PR #1146, opened 2017-09-26
Fbnil patch blocks

FBnil:
What is new?
- cloneBlock() with variables inside now expand like cloneRow() does, with #n at the end (and you can get the old behavior back with an extra parameter)
- Block functions now return sensible information instead of void
- you can throw exceptions if you can not find a block (enabled by an extra parameter)
- getBlock() implemented
- Multiple same named blocks behave as before (you can replace them one by one)
- Testcases! Now we cover a lot more (for Block functions)
- Scrutinize, phpunit and phpcs compliant now (third try, fingers crossed)

Details:
I removed the regular expression (which did not work). I got one working (fairly good at regexp), but it seems PHP7 does not like multiple non-greedy patterns. The need to anchor the query to <?xml> is very dirty). 
So I implemented the findBlockStart() and findBlockEnd(), similar to what cloneRow() uses, that search upto a paragraph change <w:p>, This has been tested with LibreOffice generated docx (that features <w:p>) and real MSOffice documents (that features extra parameters, nb: <w:p w:rsidR="00AC46F7" w:rsidRDefault="00AC46F7" w:rsidP="00AC46F7">). As well as a few new testcases to cover the new functionality.

Future: getRow(), cloneRow($search, $numberOfClones, $incrementVariables = true)
setImage() (replaces only the file inside the zip, not the formatting)

Todo: Tested under PHP 7.0.22 maybe try older versions too (although no PHP7-only have been used).

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -321,60 +321,101 @@ class TemplateProcessor
      * @param string $blockname
      * @param integer $clones
      * @param boolean $replace
+     * @param boolean $incrementVariables
+     * @param boolean $throwexception
      *
      * @return string|null
      */
-    public function cloneBlock($blockname, $clones = 1, $replace = true)
+    public function cloneBlock($blockname, $clones = 1, $replace = true, $incrementVariables = true, $throwexception = false)
     {
-        $xmlBlock = null;
-        preg_match(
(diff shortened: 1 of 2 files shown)


FBnil:
some Scrutinizer's errors left... back to the drawingboard

Issue #1121, opened 2017-08-14
phpword v0.13.1 corrupting docx file after saving


Zxurian:
( This was all working with v0.12, however we upgraded our server to PHP 7, and as a result, had to update phpword to v0.13. Since we use a later Zend Framework, I also had to fork and modify the phpword repo to not require specific `zendframework/zendframework: 2.4` but instead `zendframework/zendframework: ^2` )

I have a .docx file that I've created that has several `${xxx}` phpword placeholders in it. With the upgrade to phpword v0.13, saving the files is now corrupting them. Trying to open the document complains of a mismatched tag (xml below), and even after correcting the mismatch, the document itself still has malformed formatting from the original.

Code for template variable swapping:
```php
        $template = new PhpWord\TemplateProcessor($this->templatePath);
        foreach ($this->transposeVars() as $key => $value) {
            $template->setValue($key, $value);
        }
        $template->saveAs($fileName);
```

Original `document.xml`
```xml
[...]

PR #1147, opened 2017-09-26
FBnil patch blocks

FBnil:
I removed the regexp (although I had a working regexp, it seems PHP7 does not like multiple non-greedy patterns. The need to anchor the query to <?xml> is very dirty). 
So I implemented the findBlockStart() and findBlockEnd() that search upto a paragraph change <w:p>, This has been tested with LibreOffice generated docx (that features <w:p>) and real MSOffice documents (that features extra parameters, nb: <w:p w:rsidR="00AC46F7" w:rsidRDefault="00AC46F7" w:rsidP="00AC46F7">). As well as a few new testcases to cover the new functionality.

What is new?

- Blocks with variables inside now expand like Rows, with #n at the end (and you can get the old behavior back with an extra parameter)
- Block functions now use the same method as Row functions to detect start/end
- Block functions now return sensible information instead of void
- You can optionally throw exceptions if you can not find a block (enabled by an extra function parameter)
- getBlock() implemented
- Complete coverage of unit tests (including obvious missing tests) for Block functions
- Multiple same named blocks behave as before (you can replace them one by one)
- Went through phpcs and phpunit to test everything. All tests pass* All formatting passes*

Files modified:
./src/PhpWord/TemplateProcessor.php
./tests/PhpWord/OpenTemplateProcessor.php
./tests/PhpWord/TemplateProcessorTest.php

OpenTemplateProcessor is a subclass that allows access to the private fields in TemplateProcessor. This is required for good testing of the output results.

Tested with:
- phpunit --bootstrap autoload.php phpoffice/PHPWord/tests/PhpWord/TemplateProcessorTest.php
- phpcs ./src/PhpWord/TemplateProcessor.php ./tests/PhpWord/TemplateProcessorTest.php --standard=PSR2

Todo: Tested under PHP 7.0.22 maybe try older versions too (although no PHP7-only have been used).

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -665,7 +665,7 @@ class TemplateProcessor
         }
         if (!$tagStart) {
             if ($throwexception) {
-                throw new Exception('Can not find the start position of the row to clone.');
+                throw new Exception('Can not find the start position of the item to clone.');
             } else {
                 return 0;
             }
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -321,60 +321,101 @@ class TemplateProcessor
      * @param string $blockname
      * @param integer $clones
      * @param boolean $replace
+     * @param boolean $incrementVariables
+     * @param boolean $throwexception
      *
      * @return string|null
      */
-    public function cloneBlock($blockname, $clones = 1, $replace = true)
+    public function cloneBlock($blockname, $clones = 1, $replace = true, $incrementVariables = true, $throwexception = false)
     {
-        $xmlBlock = null;
-        preg_match(
-            '/(<\?xml.*)(<w:p.*>\${' . $blockname . '}<\/w:.*?p>)(.*)(<w:p.*\${\/' . $blockname . '}<\/w:.*?p>)/is',
-            $this->tempDocumentMainPart,
-            $matches
-        );
+        $S_search = '${'  . $blockname . '}';
+        $E_search = '${/' . $blockname . '}';
 
-        if (isset($matches[3])) {
-            $xmlBlock = $matches[3];
-            $cloned = array();
+        $S_tagPos = strpos($this->tempDocumentMainPart, $S_search);
+        $E_tagPos = strpos($this->tempDocumentMainPart, $E_search, $S_tagPos);
+        if (!$S_tagPos || !$E_tagPos) {
+            if($throwexception)
+                throw new Exception("Can not find block '$blockname', template variable not found or variable contains markup.");
+            else
+                return null; # Block not found, return null
+        }
+
+        $S_blockStart = $this->findBlockStart($S_tagPos);
+        $S_blockEnd = $this->findBlockEnd($S_tagPos);
+        #$xmlStart = $this->getSlice($S_blockStart, $S_blockEnd);
+
+        $E_blockStart = $this->findBlockStart($E_tagPos);
+        $E_blockEnd = $this->findBlockEnd($E_tagPos);
+        #$xmlEnd = $this->getSlice($E_blockStart, $E_blockEnd);
+        
+        $xmlBlock = $this->getSlice($S_blockEnd, $E_blockStart);
+
+        if($replace){
+            $result = $this->getSlice(0, $S_blockStart);
             for ($i = 1; $i <= $clones; $i++) {
-                $cloned[] = $xmlBlock;
+                if($incrementVariables)
+                    $result .= preg_replace('/\$\{(.*?)\}/', '\${\\1#' . $i . '}', $xmlBlock);
+                else
+                    $result .= $xmlBlock;
             }
+            $result .= $this->getSlice($E_blockEnd);
 
-            if ($replace) {
-                $this->tempDocumentMainPart = str_replace(
-                    $matches[2] . $matches[3] . $matches[4],
-                    implode('', $cloned),
-                    $this->tempDocumentMainPart
-                );
-            }
+            $this->tempDocumentMainPart = $result;
         }
 
         return $xmlBlock;
     }
 
     /**
+     * Get a block. (first block found)
+     *
+     * @param string $blockname
+     * @param boolean $throwexception
+     *
+     * @return string|null
+     */
+    public function getBlock($blockname, $throwexception = false){
+        return $this->cloneBlock($blockname, 1, false, $throwexception);
+    }
+    /**
      * Replace a block.
      *
      * @param string $blockname
      * @param string $replacement
+     * @param boolean $throwexception
      *
-     * @return void
+     * @return false on no replacement, true on replacement
      */
-    public function replaceBlock($blockname, $replacement)
+    public function replaceBlock($blockname, $replacement, $throwexception = false)
     {
-        preg_match(
-            '/(<\?xml.*)(<w:p.*>\${' . $blockname . '}<\/w:.*?p>)(.*)(<w:p.*\${\/' . $blockname . '}<\/w:.*?p>)/is',
-            $this->tempDocumentMainPart,
-            $matches
-        );
+        $S_search = '${'  . $blockname . '}';
+        $E_search = '${/' . $blockname . '}';
 
-        if (isset($matches[3])) {
-            $this->tempDocumentMainPart = str_replace(
-                $matches[2] . $matches[3] . $matches[4],
-                $replacement,
-                $this->tempDocumentMainPart
-            );
+        $S_tagPos = strpos($this->tempDocumentMainPart, $S_search);
+        $E_tagPos = strpos($this->tempDocumentMainPart, $E_search, $S_tagPos);
+        if (!$S_tagPos || !$E_tagPos) {
+            if($throwexception)
+                throw new Exception("Can not find block '$blockname', template variable not found or variable contains markup.");
+            else return false;
         }
+
+        $S_blockStart = $this->findBlockStart($S_tagPos);
+        $S_blockEnd = $this->findBlockEnd($S_tagPos);
+        #$xmlStart = $this->getSlice($S_blockStart, $S_blockEnd);
+
+        $E_blockStart = $this->findBlockStart($E_tagPos);
+        $E_blockEnd = $this->findBlockEnd($E_tagPos);
+        #$xmlEnd = $this->getSlice($E_blockStart, $E_blockEnd);
+        
+        $xmlBlock = $this->getSlice($S_blockEnd, $E_blockStart);
+        
+        $result  = $this->getSlice(0, $S_blockStart);
+        $result .= $replacement;
+        $result .= $this->getSlice($E_blockEnd);
+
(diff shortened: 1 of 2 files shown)


FBnil:
Not sure what to do with the low scores of Scrutinizer. I can roll back the duplication of code, and use less functions-in-functions? (at a cost of more duplicate code, get it to B and C instead of D)

FBnil:
Updates on 2017-10-02

1. Single-Element Blocks `${blockname/}`

You know the problem with adding lists? You need to put a block before and after one numbered list like so:
```
${myblock}
* ${myelement}
${/myblock}
```
This then can be used with cloneBlock('myblock',2) and setValue('myelement#1', 'first substitution') etc.

Now we allow this:

```
* ${myelement/}
```
And we use: cloneBlock('myelement/',2) and setValue('myelement/#1', 'first substitution') etc.
Which is more succint and easier on the eyes in a template.

2. Update files inside the working zip file.

We also added zipAddFromString(), which works like this if you want to replace an image in the docx, while keeping the position and size of that image in the document:

```
	$file_dir = storage_path('document_with_image.docx');
	$templateProcessor = new \PhpOffice\PhpWord\TemplateProcessor($file_dir);
	$replace = "word/media/image2.jpeg";
	$met = file_get_contents(storage_path('big.jpg'));
	$result = $templateProcessor->zipAddFromString($replace,$met);	
	$templateProcessor->saveAs(storage_path('app/public/x.docx'));
	return response()->download(storage_path('app/public/x.docx'))->deleteFileAfterSend(true); 
```

So  `"word/media/image2.jpeg"` is the name of the image inside the docx/zip file. And `$met` is the binary content of the new image. The resulting file was tested on Office 365 and LibreOffice and works as expected.

3. Multi line values
Also, and even though experimental, it seems to work quite well, as it has been proposed by other people and they came up with the replacement code. It allows for:
`->setValue("var", "my \n multiline \n string");`
Which is the equivalent to shift-Enter while writing a document (stay on the same bulletpoint/tablecell, but allow multiple lines).

drjonnicholson:
Would be interested to know if this is going to be merged?

FBnil:
I am doing my utmost (code quality wise and helping out) to get noticed and get my code merged in. However, there are many other changes in line (for example, https://github.com/PHPOffice/PHPWord/pull/1084 is great and has been waiting 4 months, https://github.com/PHPOffice/PHPWord/pull/779 has been waiting more than a year, etc), and the people that have access to the PHPWord are, uhm, scarce. I only see @troosan active (doing only bugfixes, not new functionality?).

Meanwhile, if you need the functionality because of PHP7, I only changed 1 file (TemplateProcessor, maybe Styles) so you can just drop those in (replace the files manually in a fresh 0.13.0). I'll be around 'till december or so. Offering support, and slowly adding documentation for the new functionality (because, hey, gotta work on the website I'm building too).

drjonnicholson:
Thanks for the update @FBnil. Open source is open source, hopefully the PHPOffice team will have capacity to go through these in the not too distant future.

At the moment we are working around the issue by generating a template from a meta template using a blocks of XML code copied out of a sample document offline - basically ignoring the blocks functionality and inserting raw code into text line templates. It's a bit of a kludge, but it gets the job done for now.

FBnil:
Added rudimentary support for inline-blocks.
```
So now ${inline} we can ${/inline} use this.
cloneBlock('inline', 'you can');
```

FBnil:
added setBlock(). A closed block ends with slash and is treated as both a replacement variable AND a paragraph block, so, you can write:
```
This is the start of a paragraph, and ${variable/} is in this. 
```
`cloneBlock('variable/', 2)` already cloned the block:
```
This is the start of a paragraph, and ${variable/#1} is in this. 
This is the start of a paragraph, and ${variable/#2} is in this. 
```
and you can `setValue('variable/#1', "rum"); setValue('variable/#1', "cola");`

Now you can do this in one go: `setBlock('variable/', "rum\ncola", 1);`

It also accepts arrays, like `setValue()`

FBnil:
Exteriorize a global bool TemplateProcessor::$ensureMacroCompletion, so that you can disable macrocompletion (pulling everything that looks like `$`, `{` , `TEXT` and `}` together). 
This keeps your document xml untouched, and you can then search for words, for example, for censorship reasons:

```
TemplateProcessor::$ensureMacroCompletion = false;
setValue( 'butt', 'gluteus maximus' );
setValue( 'dictatorship', 'benevolent' );
```

FBnil:
Well, thanks to Scrutinizer I was pointed to a 1 year old bug: With `setValue()` you can replace macro tags in the main document, but also Header and Footers. Due to the bug, only the first header and first footer (because those are actually arrays, not strings) were modified. Which left me wondering if I had made the same mistake and yes. Segments were also faulty. So segments now can be called like this:
```
$templateProcessor->cloneSegment('${headerValue}', 'w:p', 'Headers:1', 2);
$templateProcessor->cloneSegment('${documentContent}', 'w:p', 'MainPart', 2);
$templateProcessor->cloneSegment('${footerValue}', 'w:p', 'Footers:1', 2); 
```
The first footer is `Footers:1` there is no zero.
So let's see how that affects the code quality score...

FBnil:
Ok, I removed makeTable() and toXML(), in the hope to find time and add that to another function.
This moves the Quality of the TemplateProcessor() class back from F to D. 

I also tackled the problem that fixBrokenMacros() has in b0rking a document at load time due to runaway macro's using strip_tags() to remove all xml. We set up a few rules:

1. a macro can not have `<`,`>`,`{`or`}` in it

This is reasonable, we only use text, some have extended tags with `:` and `,` in it, which still work.

2. Macro's must be defined in the same paragraph and can not span nested paragraphs.

This too is reasonable because visually, you always keep the macro together. However, in order to accomplish this, I used preg_split and implode(). As you know, those are very expensive functions when using big documents. (because there are a lot of `<w:p>` tags!)
I tried other ways, to use PREG_OFFSET_CAPTURE and substrings, to use xpath `//w:p` to to the same, but guess what, those actually xml cut and build the whole document (slow), and are actually ok when manipulating it... but the code was much larger, so I stopped with that. 

So the method is cut at `</w:p>` and `<w:p>`, then inside each block to the magic it was doing in the past. This should be tested in the field now. Although I am confident, as the tests succeed.  

Speaking of tests, there is still the problem I use eval to generate a TemplaceProcessor with accessible private functions and variables (very handy for testing)... but Travis doesn't like it. Does somebody know how I can fix that? (other constraints: I can not call evaluate_once, I can not use a class in another file.)

I also tried to optimize findTagLeft(), but the code written for *Row is as efficient as what I did (looking for an open tag, say `<w:p`) and then check if '>' or ' ' comes after it, keep checking if not. It looped about 4 times (finding w:pPr and those kind of tags). Which is the same time that strrpos() takes to look for the first tag up onto the beginning of a document, then try the other tag ... in less code.

FBnil:
Today I discovered Reflections.  So I removed my clutch that broke Travis.

FBnil:
Well, all the block and segment logic has been moved to process* functions, and (clone|get|replace|delete)* functions now call that function. Still 3 functions to go for 100% coverage and then I rest.
Meanwhile:  *Row functions have been fixed when using docx generated by the latest LibreOffice. (MSOffice adds  extra spaces to a closed tag, and this was the default detection)

FBnil:
What is so great about these static changes? Well, you can make a subclass where your macro is defined as `%{macro}` and have a few of the core functions redefined to automatically handle the change. Or you can rewrite a few more functions to have default values in your macro's but still able to call the setValue by the short name (without the extra user definable part). I have working examples, still need to put it somewhere...

FBnil:
Received confirmation it closes https://github.com/PHPOffice/PHPWord/issues/1121

FBnil:
Well, I can not get rid of the conflict, but I guess TemplateProcessor has been rewritten and is now conflicting with the parameters the old TemplateProcessor had (and I've added quite a few new parameters).

Ok, latest additions: Segments now allow to search left or right. So take a needle, find that position, in the past, we only could search around it, that is, start tag was to the left and end tag was to the right, now we can search for both start/end tags to the right or left. 
For this new functionality, I had to add more code than I wanted, and of course, it was buggy. But now I just added new testcases and they now pass. So that is one bug less.

The rest of the functional additions will be through traits or subclasses.

troosan:
@FBnil do you think you could come up with some proper unit testing for the TemplateProcessor.
Maybe we can dedicate a release (0.15) to fixing/improving this part of the code.

FBnil:
Will try. It seems only space related problems that Travis is giving. Give me a few days.

FBnil:
@troosan Just to be clear I understand you: Testcases for the current TemplateProcessor in dev-master ? And with "proper", that includes reflections and maybe mocking?

troosan:
There currently are some tests defined already (in TemplateProcessorTest) but I have a feeling a lot of cases are not covered by those tests. That's the reason why we have so many issues with it. I would like to have a reasonable assurance that when we touch this class we do not break anything else. Today I really am not.

MyDigitalLife:
I just tested there changes in my document where i'm having problems with cloning docs, using php 7.1. Looks like this definitely solved my problems. Keep up the good work! I hope this will be merged and released soon!

FBnil:
I installed the fixer with `composer require friendsofphp/php-cs-fixer`
and tested then removed `--dry-run` to let the fixer automatically update the code:
`php-cs-fixer fix --diff --verbose --dry-run src/PhpWord/TemplateProcessor.php`
Glad it is green again (Scrutinizer & Travis). 

My testing script is now:
```BASH
PROJECT=/var/www/html/cvprima
cd $PROJECT/vendor
cp phpoffice/PHPWord/src/PhpWord/TemplateProcessor.php phpoffice/phpword/src/PhpWord/TemplateProcessor.php;
phpunit -c $PROJECT/vendor/phpoffice/PHPWord/phpunit.xml.dist --bootstrap autoload.php phpoffice/PHPWord/tests/PhpWord/TemplateProcessorTest.php   --coverage-html $PROJECT/public/coverage
cd -

cd /home/prima/github/PHPWord
phpcs ./src/PhpWord/TemplateProcessor.php ./tests/PhpWord/TemplateProcessorTest.php --standard=PSR2  --colors -v
phpmd src/,tests/ text ./phpmd.xml.dist --exclude pclzip.lib.php
phpdoc -q -d ./src -t ./build/docs --ignore "*/src/PhpWord/Shared/*/*" --template="responsive-twig"
php-cs-fixer fix --diff --verbose --dry-run ./src/PhpWord/TemplateProcessor.php
php-cs-fixer fix --diff --verbose --dry-run ./tests/PhpWord/TemplateProcessorTest.php
cd -
```



Now looking into making more testcases...

MyDigitalLife:
@FBnil @troosan Any idea when this will be merged?

troosan:
@MyDigitalLife I'd like to have a release (0.15?) dedicated to merge the different fixes/improvements on the TemplateProcessor, I guess this will need a bit of work :-P

MyDigitalLife:
@troosan Could you create a milestone for 0.16 and add this then? If there are any other issues that need to be looked at I can give it a go.

PR #1161, opened 2017-10-19
Get rid of duplicated code in TemplateProcessor.php

abcdmitry:


FBnil:
Why not `self::ensureMacroCompleted` like in the function `setValue`?

abcdmitry:
This is a late static binding feature allowing to overwrite `ensureMacroCompleted` method from the child class.
Right now I work on extending this class within my project and I need to overwrite this method.

Actually replacing `self::` with `static::` within the `setValue()` would be nice as well, since it doesn't affect current functionality, but offers flexibility for those willing to extend the class.

FBnil:
Aha.. I did a complete rewrite of  [TemplateProcessor](https://github.com/FBnil/PHPWord/blob/4f653336bdd93544422807a2d1f2a089cdd7d1a7/src/PhpWord/TemplateProcessor.php) .
Maybe this will be useful:
```
Enable/disable setValue('key') becoming setValue('${key}') automatically.
Call it like: TemplateProcessor::$ensureMacroCompletion = false;
```

FBnil:
While you are at cleanups, do not forget:

setValue():

```
        $this->tempDocumentMainPart = $this->setValueForPart($search, $replace, $this->tempDocumentMainPart, $limit);

        $documentParts = array('tempDocumentHeaders', 'tempDocumentFooters');
        foreach ($documentParts as $tempDocumentParts) {
            foreach ($this->{$tempDocumentParts} as &$tempDocumentPart) {
                $tempDocumentPart = $this->setValueForPart($search, $replace, $tempDocumentPart, $limit);
            }
        }
```
Because the Headers and Footers are arrays.

I did not bother too much with unset, the current code does not bite itself, and when the function ends, they are out of scope and cleaned up automatically. It IS, however, good practice.

abcdmitry:
I am not sure where the code `foreach ($this->{$tempDocumentParts} as &$tempDocumentPart) {` is located at. It is not in the `develop`'s branch `TemplateProcessor.php` file.
In my corrections at #1162 there is one place where two `foreach`'es with `&$item` go one after another. So it is definitely a problem.

FBnil:
I mean that PHP as a language is not THAT broken, this works perfectly fine:
```
<?php

$Q = [2,4,6];
$R = [10,20,30];

foreach($Q as &$item){
        $item++;
}

foreach($R as &$item){
        $item++;
}

var_dump($Q);
var_dump($R);
```
Prints [3,5,7] and [11,21,31], as expected

Granted, at the end of both and within the same function, you should never use item, because:
```
$item = 100;
var_dump($R);
```
(Prints [11,21,**100**])

But in our case, we are in a function, and when the function ends, the variable goes out of scope and all is fine:
```
<?php

$Q = [2,4,6];
$R = [10,20,30];

function addOne(&$Arr){
        foreach($Arr as &$item){
                $item++;
        }
}

addOne($Q);
addOne($R);
$item = 100;

var_dump($Q);
var_dump($R);
```

FBnil:
FYI:  [Existing bugs in TemplateProcessor](https://scrutinizer-ci.com/g/PHPOffice/PHPWord/inspections/de812839-ed05-4e0b-ba62-9c7989befb47/issues/files/src/PhpWord/TemplateProcessor.php?status=all&orderField=path&order=asc&honorSelectedPaths=0)

I've ironed them out in my branch.

PR #1166, opened 2017-10-21
Fixed broken lists

Piterden:


troosan:
Thanks for this but the formatting was already corrected on the current development branch

PR #1168, opened 2017-10-22
Html::addHTML should parse <b> <i> <u>

FBnil:
**Abstract:** 

Allows for the HTML import of the following:

- `<span style="font-weight: bold;">` as `<w:b/>`
- `<span style="font-style: italic;">` as `<w:i/><w:iCs/>`
- `<b>` as `<w:b/>`
- `<i>` as `<w:i/><w:iCs/>`

**Motivation:**

A few years ago there was a mindset to drop b and i and replace them with strong/em. But these have not gone away and thus, we need to support these. Now, if you look at the code: `"./src/PhpWord/Shared/Html.php"`

```
 'strong'    => array('Property',    null,   null,       $styles,    null,   'bold',         true),
'em'        => array('Property',    null,   null,       $styles,    null,   'italic',       true),
```
It seems foolish to me to say that strong (which conveys semantic meaning) maps to.... "bold" and em (which conveys semantic meaning) maps to.... "italic". They do not, in a W3C strict kind of sense. But if they did (as lot of software does)... then "b" and "i" are the real inheritors. Indeed, in OOXML bold maps to `w:b` not `w:strong` and we have `<w:i/>` and `<w:iCs/>` too. Very logical `b -> w:b` and `i -> w:i`

**Application:**
Let's say you are using the TemplateProcessor to replace some text, but the text is made with `bootstrap-wysiwyg` or similar tools that spews html. No fear! We can do this now (we always could, but required a xslt transformation to use em and strong):

```
use PhpOffice\PhpWord\Shared\Html as HTMLParser;

	$htmlStr = '<html><body><p><strong>hello</strong><i>cool</i> world</p><p><span style="font-style: italic;">PHP</span> is <span style="font-weight: bold;">AWESOME</span></p></body></html>';
	HTMLParser::addHtml($section, $htmlStr, true);
	$fullXml = $objWriter->getWriterPart('Document')->write();
	dd($fullXml);  // or use var_dump()
```
Now cut the $fullXml at the right place, and use replaceBlock()... now doesn't that feel nice...

**Notes**
`addHtml()` needs well formed html, this often means you need to wrap your string between: `<html><body><p>` and `</p></body></html>` or the parser will die (so for a robust product you must use try-catch).

I tried to use inlineStyle directly, as that would mean less code, but the style we want to set is in `$style['font']` so that did not work. I also thought it was related to `parseProperty()`, although a separate method could have been cleaner. If you are the maintainer and want such changes... sure. 

A shout out to @fr05t1k for showing me this part of the code, and to the issue raisers telling us that HTML lack functionality.

**PHPWord Maintainers**
Testcases? *laughs nervously* Of course I did test it with nested properties and what not. Uhmm... if you give me a commitment this is merged in, I'll write them. With the TemplateProcessor rewrite I took around 8 times the amount of time to add testcases (granted, I was aiming for 100%)


troosan:
@FBnil I added a couple of unit tests.
One important missing feature is the import of HTML tables which does not work.

Issue #1173, opened 2017-10-26
read docx error when contains image from remote url


yeluolei:
I have a docx with image from an url, the rels like this
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Relationships 
    xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
    <Relationship Id="rId3" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/webSettings" Target="webSettings.xml"/>
    <Relationship Id="rId7" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/theme" Target="theme/theme1.xml"/>
    <Relationship Id="rId2" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/settings" Target="settings.xml"/>
    <Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/styles" Target="styles.xml"/>
    <Relationship Id="rId6" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/fontTable" Target="fontTable.xml"/>
    <Relationship Id="rId5" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/image" Target="https://image0.lietou-static.com/normal/5773448845ce6e5083b716f404a.png" TargetMode="External"/>
</Relationships>
```

but when I try to load this docx to php, it throw an exception like this 

```
Exception 'PhpOffice\PhpWord\Exception\InvalidImageException' with message ''

in /app/vendor/phpoffice/phpword/src/PhpWord/Element/Image.php:386

Stack trace:
#0 /app/vendor/phpoffice/phpword/src/PhpWord/Element/Image.php(139): PhpOffice\PhpWord\Element\Image->checkImage('zip://liepin_1....')
#1 [internal function]: PhpOffice\PhpWord\Element\Image->__construct('zip://liepin_1....')
#2 /app/vendor/phpoffice/phpword/src/PhpWord/Element/AbstractContainer.php(145): ReflectionClass->newInstanceArgs(Array)
#3 [internal function]: PhpOffice\PhpWord\Element\AbstractContainer->addElement('Image', 'zip://liepin_1....')
#4 /app/vendor/phpoffice/phpword/src/PhpWord/Element/AbstractContainer.php(112): call_user_func_array(Array, Array)
```

seams it try to unzip the remote url as image data and failed

FBnil:
[...]

PR #1176, opened 2017-10-27
Allow reading of TargetMode for external images

FBnil:
Abstract: 
    This patch changes the reader in order to handle TargetMode="External"
Closes: 
    https://github.com/PHPOffice/PHPWord/issues/1173

Explanation:

When reading from documents.xml.rels, we not often see external images, like this:
`<Relationship Id="rId4" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/image" Target="https://somewhere/image.png" TargetMode="External"/>`

The TargetMode needs to be read and understood, to NOT default to a ZIP container for type "image" if the image is external.
Failing to do so results in an InvalidImageException while reading the file.


troosan:
@FBnil thanks for this, I manually did (almost) the same changes to avoid merge issues.

PR #1185, opened 2017-11-08
rename 'Object' classes to 'ObjectElement' (php 7.2 compatibility)

SailorMax:
If name is not very good, now easier to find all require places to rename ;)


SailorMax:
Scrutinizer said "Failed", but I can't see what's the problem?

troosan:
@SailorMax not a big fan of the name ObjectElement, it is too different from all other elements.
What about OLEObject?

SailorMax:
I think you better know.
I don't know internal structure of PHPWord and used `ObjectElement` as name similar to `AbstractElement`.
If it looks like `OLEObject`, you can rename it :)

troosan:
still need https://github.com/tecnickcom/TCPDF/pull/74 to be fully PHP 7.2 compatible

stevechilds:
@troosan - Has your fix been pushed into the main branches yet as the 7.2 "Object" class bug is causing us issues on our new Ubuntu 18.04 setup, which uses PHP 7.2 ?

I've tried 0.14, which is what we were on, which fails and even using "phpoffice/phpword": "dev-master" failed to resolve the issue.

PR #1213, opened 2017-12-01
Test against PHP 7.2

carusogabriel:
# Description

With the release of [`PHP 7.2`](http://php.net/archive/2017.php#id2017-11-30-1), would be interesting to test it.

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have run phpunit, phpcs, php-cs-fixer, phpmd
- [ ] The new code is covered by unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ php:
     - 5.6
     - 7.0
     - 7.1
+    - 7.2
 
 matrix:
     include:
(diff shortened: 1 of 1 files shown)


troosan:
I added the PHP 7.2 build to #1185, which fixes a blocking error in PHP 7.2, which will have to correct anyway if we want to be 7.2 compatible.

PR #1221, opened 2017-12-13
correctly parse on/off values (w:val="true|false|1|0|on|off")

### LINKED ISSUE #1219: Incorrect definition of bold text
`<w:b w:val="0"/>` don't display as bold text in MSWord, but when reading from docx file `Style\Font::isBold()` is `true`

* PHP version: 7.1.11
* PHPWord version: 0.14
screenshot

troosan:
# Description
The value of w:val of type OnOff can be any of http://www.datypic.com/sc/ooxml/t-w_ST_OnOff.html
0, off and false should be treated as false

Fixes #1219 

# Checklist:

- [x] I have run phpunit, phpcs, php-cs-fixer, phpmd
- [x] The new code is covered by unit tests


FBnil:
Thanks troosan. It's these little things that make the PHPWord better

PR #1258, opened 2018-01-25
Add parsing of HTML links

### LINKED ISSUE #1255: Add links inside HTML paragraphs
`addHtml` function converts my HTML, but strips any `a href` links.

What would be the best way to convert the HTML while keeping the links inside paragraphs and other HTML elements?

troosan:
### Description

Add parsing of `<a href="">`

Fixes #1255 

### Checklist:

- [x] I have run `composer check` and no errors were reported
- [x] The new code is covered by unit tests


rickydeez:
Awesome work, thanks!

Anyway to style the links inside HTML, besides any inline CSS?

troosan:
I tried with `style="text-decoration: underline"` but somehow it's not being parsed correctly.
If you find a solution I'm happy to merge it :-)

Issue #664, opened 2015-11-20
TemplateProcessor cloneBlock breaks my document


btry:
Hi

This is my first use of PHPWord. I created a docx file with LibreOffice (do not ask why I'm not using MS Office :) )

The document contains a block (a single paragraph) to be cloned. The resulting docx file appears to be empty (still using LibreOffice).

I noticed the paragraphs are different from the sample_23 provided by PHPWord and those in my document created by LibreOffice. 

The following code has been code formated (with Eclipse) to check the XML structure.

Sample_23 provided by PHPWord repository (block tag only): 
```
		<w:p w:rsidR="00C0566D" w:rsidRPr="003B08B6" w:rsidRDefault="00C0566D"
			w:rsidP="00C0566D">
			<w:r>
				<w:t>${CLONEME}</w:t>
			</w:r>
		</w:p>
```

Paragraph generated by LibreOffice (block tag only)
```
		<w:p>
			<w:pPr>
				<w:pStyle w:val="Corpsdetexte" />
				<w:pageBreakBefore w:val="false" />
				<w:rPr></w:rPr>
			</w:pPr>
			<w:r>
				<w:rPr></w:rPr>
				<w:t>${itemtypeBlock}</w:t>
			</w:r>
		</w:p>
```

output file (cloned blocks appears to be nested)
```
		<w:p>
			<w:pPr>
				<w:pStyle w:val="Corpsdetexte" />
				<w:p>
					<w:pPr>
						<w:pStyle w:val="Corpsdetexte" />
						<w:rPr></w:rPr>
					</w:pPr>
					<w:r>
						<w:rPr></w:rPr>
						<w:t>cloned paragraph</w:t>
					</w:r>
				</w:p>
				<w:p>
					<w:pPr>
						<w:p>
							<w:pPr>
								<w:pStyle w:val="Corpsdetexte" />
								<w:rPr></w:rPr>
							</w:pPr>
							<w:r>
								<w:rPr></w:rPr>
								<w:t>cloned paragraph</w:t>
							</w:r>
						</w:p>
						<w:p>
							<w:pPr>
								<w:p>
									<w:pPr>
										<w:pStyle w:val="Normal" />
										<w:rPr></w:rPr>
									</w:pPr>

```

I began to debug and found the following : 
- the XML tree is broken in the output file (cloned blocks nested, and some XML tags are not properly closed

btry:
I finally found a working regex on my document. It has been designed to run also with the sample_23 document.

[...]

PR #1269, opened 2018-02-02
Fix regex in `cloneBlock` function

### LINKED ISSUE #681: unreadable file content generated using template processor
Hi, my problem is with using the template processor class. 

I have a 1x3 table in a template doc file. In one of the columns I have a block of text that I wish to clone or delete using cloneBlock or deleteBlock function. I notice that without styling cloneBlock and deleteBlock functions work as expected, but if I changed the font color of the block of text to white the generated doc file cannot be read on my computer with the following error message:

The Open XML file BasicTable(40).docx cannot be opened because there are problems with the contents or the file name might contain invalid characters(for example, V)...No error detail available.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28886944-unreadable-file-content-generated-using-template-processor?utm_campaign=plugin&utm_content=tracker%2F323108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F323108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nicoder:
it wrongly matched `<w:pPr` when looking for `<w:p`

in that case the cloned block would not be valid XML,
and so the generated docx document could not be opened by Word.

adding the `\b` word boundary to the regex fixes this.

the accompanying unit test failed with this message when trying to load
the generated file:

```
ErrorException: DOMDocument::loadXML(): Opening and ending tag mismatch:
p line 2 and body in Entity, line: 2
```

Fixes #681 and #664.


nicoder:
Hi,

If / when this is merged I would like to open an other pull request to teach `cloneBlock` to only clone some of the blocks with that name.
    
By giving it the same `$limit` parameter as `setValue` has.

czosel:
Hi @nicoder,
nice to see a familiar face 😉
I’ve been refactoring the Regex-based approach to use DOMDocument in #1354 - I‘d really appreciate your feedback!

PR #841, opened 2016-07-25
add support for table layout

aoloe:
optionally avoid the table to stretch to the page width (only for word output)

PR #1276, opened 2018-02-09
Added support for fixed tableLayout, allowing to use fixed width columns in tables

ekopach:
Added the function to fix the width of columns in tables, when text very long and without delimiters.

- [x] I have run `composer check` and no errors were reported
- [ ] The new code is covered by unit tests

### DIFF
--- a/src/PhpWord/Style/Table.php
+++ b/src/PhpWord/Style/Table.php
@@ -591,7 +591,7 @@ class Table extends Border
 
     public function setFixedColsWidth($value = true)
     {
-        $this->cantSplit = $this->setBoolVal($value, $this->fixedColsWidth);
+        $this->fixedColsWidth = $this->setBoolVal($value, $this->fixedColsWidth);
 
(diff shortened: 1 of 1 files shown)


troosan:
@ekopach actually, this is already implemented in #841, which has not yet been merged.

ekopach:
Oh, thanks. Sorry, I have not seen.

samimussbach:
@troosan I could test and review one of the PRs as we have the same issue. Which one do you prefer?

troosan:
@samimussbach I don't really have a preference. This solution seems simpler, as only 2 values are possible and 1 the default value, I guess just a boolean is enough.
Both PR are missing unit test, doc, ... though :-S
It would also be nice to immediately implement the reader for it.

troosan:
@samimussbach please check out the following #1281

samimussbach:
You where faster than me, nice. Will check.

Issue #1256, opened 2018-01-23
Zend\Stdlib dependency?


JamesB7:
This is:

- [ ] a bug report
- [X] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpword)

Could you please consider removing the Zend\Stdlib dependency? It brings many classes, some of which unserialize data, into the autoloader. I would like to avoid increasing the attack surface of my program so extremely. PHPWord is only dependent on one single-line function in the whole library.

Thank you!

James

PR #1284, opened 2018-02-15
Remove zend-stdlib dependency

Trainmaster:
### Description

Fixes https://github.com/PHPOffice/PHPWord/issues/1256

### Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,6 @@
         "php": "^5.3.3 || ^7.0",
         "ext-xml": "*",
         "zendframework/zend-escaper": "^2.2",
-        "zendframework/zend-stdlib": "^2.2 || ^3.0",
         "phpoffice/common": "^0.2"
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


troosan:
@Trainmaster a better solution would be to use PhpOffice\Common\Text:: isUTF8 instead.
I'll see if can un-deprecate it (@Progi1984?).

Progi1984:
👌 with that.

Trainmaster:
@troosan I changed it to `PhpOffice\Common\Text:: isUTF8`.

JamesB7:
I think the is_string check is important: preg_match will return true for non-strings, such as integers. (Yes, this surprised me too. Hurrah for PHP?)

Without the is_string check, callers of isUTF8 (such as toUTF8) may pass through non-string values without converting them to string (such as utf8_encode does).

troosan:
I create a PR in https://github.com/PHPOffice/Common/pull/19 to modify the method

Progi1984:
Merged ;)

PR #1285, opened 2018-02-15
Improve attribute value of On/Off type

Trainmaster:
### Description

According to http://www.datypic.com/sc/ooxml/t-w_ST_OnOff.html valid values for `w:ST_OnOff` are:
- true/false (<- the pair I chose)
- on/off
- 0/1

### Checklist:

### DIFF
--- a/src/PhpWord/Writer/Word2007/Part/Settings.php
+++ b/src/PhpWord/Writer/Word2007/Part/Settings.php
@@ -174,11 +174,8 @@ class Settings extends AbstractPart
             return;
         }
 
-        if ($booleanValue) {
-            $this->settings[$settingName] = array('@attributes' => array());
-        } else {
-            $this->settings[$settingName] = array('@attributes' => array('w:val' => 'false'));
-        }
+        $value = $booleanValue ? 'true' : 'false';
+        $this->settings[$settingName] = array('@attributes' => array('w:val' => $value));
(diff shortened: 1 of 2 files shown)


Trainmaster:
@troosan In https://github.com/PHPOffice/PHPWord/blob/develop/src/PhpWord/Reader/Word2007/AbstractPart.php#L571 it says 

> null is considered true as it means the w:val attribute was not present

Is this behaviour defined somewhere?

troosan:
@Trainmaster no really, it's just the behaviour I saw when analysing documents produced by word. Of course I guess it does not hurt to be more explicit!

Trainmaster:
@troosan I found the following:

> A value of 1 or true specifies that the property shall be turned on. This is the default value for this attribute, and is implied when the parent element is present, but this attribute is omitted.

(§22.9.2.7 in ECMA-376 5th Edition Part 1)

That conforms with your analysis. But as you said it does not hurt to be more explicit here. Maybe I just change `true`/`false` to `1`/`0` to save a few bytes.

troosan:
I'm not sure the few bytes will do the difference, I prefer true/false (not that I usually go read the XML :-))

Trainmaster:
@troosan I just rebased this branch and also improved the assertions for the hyphenation setting.

PR #1290, opened 2018-02-18
Fix listitem parsing

troosan:
### Description
Formatting inside list items is lost when parsing word document

Fixes #594

### Checklist:


moromete:
Please also add a ListItemRun class on \Writer\HTML\Element so that the list can also be visible in HTML. I think the simple approach is to just have it extended from PhpOffice\PhpWord\Writer\HTML\Element\TextRun

PR #1295, opened 2018-02-21
Fix HTML parsing when style attribute is empty

lalop:
### Description

This fix the style parsing process when style is empty.
`list($a, $b) = $c` throw a notice if `$c` contain less than 2 entries.

### Checklist:

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -452,8 +452,10 @@ class Html
     private static function parseStyle($attribute, $styles)
     {
         $properties = explode(';', trim($attribute->value, " \t\n\r\0\x0B;"));
+        
         foreach ($properties as $property) {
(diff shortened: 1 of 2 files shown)


lalop:
There is an error in test due to tecnickcom/tcpdf, this library doesn't seems to be compatible with php 5.3 although this is not reflected in their composer.json
Anyway this library is deprecated and they don't take any patch

troosan:
Indeed, there’s already a or to fix it 
https://github.com/tecnickcom/TCPDF/pull/84

lalop:
maybe but they will not update this lib anymore according to readme 
> A new version of this library is under development at https://github.com/tecnickcom/tc-lib-pdf and as a consequence this version will not receive any additional development or support. This version should be considered obsolete, new projects should use the new version as soon it will become stable.

troosan:
indeed, but they released this version just yesterday so we can hope they will release a new one with the fix!

Issue #862, opened 2016-08-15
Tabs not working in titles


judgej:
Setting a tab in a paragraph works fine:

    $section->addText("First\tSecond\tThird", [], 'paragraphStyleWithTabs');

The First/Second/Third words get correctly positioned at the tab stops defined in the paragraph style.

When creating a title, this does not seem to work:

~~~php
            $phpWord->addTitleStyle(
                3,
                [],
                [
                    'tabs' => array(new \PhpOffice\PhpWord\Style\Tab('left', 1600)), // etc.
                ]
            );

    $section->addTitle(htmlspecialchars("First\tSecond\tThird", 3);
~~~

The resulting DOCX document puts a space into the position where the tab should be.

However, the tab starts *are* correctly added to the ruler in the title, so manually replacing the space with a tab inside MS Word takes the title text to the correct tab positions. The tabs *are* getting in there, *are* supported by MS Word, but the title text is not getting its `"\t"` tab characters handled as text tabs in the generated document.

judgej:
The same happens for Libra Office, except none of the tabs work (titles or paragraphs). The RTF output passes the raw tab character through. I'm wondering if the instance it is working for DOCX is a pure fluke, and I've just got the whole syntax wrong?

judgej:
Looking into the DOCX source, it looks like tab characters are carried through to that too, so they are in there.

judgej:
Looking at the source of the DOCX document after saving it with a manual tab added, MS Word seems to add a tab as an XML element rather than a tab character. So this text run will generate a tab between "One" and "Two":

~~~php
$textrun = $section->createTextRun('entryFirst');
$textrun->addText('One');
$textrun->addText("<w:r><w:tab/></w:r>");
$textrun->addText('Two');
$textrun->addTextBreak();
~~~

I'm not sure how to get a text run unto a heading though, without generating an invalid DOCX file.

judgej:
Spent hours trying to understand the code, but have not yet been able to wrap my head around it. What I suspect needs to happen to fix this, is text and title elements (and presumably text runs) that contain `\t` (tab) characters need to be split into multiple text elements - text elements and tab elements. So `"One\tTwo"` becomes three elements in a paragraph or title, "<w:t>One</w:t>", "<w:t><w:tab/></w:t>" and "<w:t>Two</w:t>".

judgej:
[...]

PR #1299, opened 2018-03-01
Add support for reading <w:tab/> element in runs.

zobo:
### Description

Adds support for Word2007 reader to understand <w:tab/> element in <w:r>. All samples seen so far had only either w:tab or w:t elements. Encoded as "\t" character. This will still display correctly in MS Word, however a proper writer should also be implemented.

Related to #862 .

### Checklist:

### DIFF
--- a/src/PhpWord/Reader/Word2007/AbstractPart.php
+++ b/src/PhpWord/Reader/Word2007/AbstractPart.php
@@ -241,9 +241,11 @@ abstract class AbstractPart
             if ($xmlReader->elementExists('w:br', $domNode)) {
                 $parent->addTextBreak();
             }
-            if ($xmlReader->elementExists('w:t', $domNode)) {
+            if ($xmlReader->elementExists('w:t', $domNode) || $xmlReader->elementExists('w:tab', $domNode)) {
                 // TextRun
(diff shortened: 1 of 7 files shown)


troosan:
Hi @zobo `w:tab` is already supported (`PhpOffice\PhpWord\Style\Tab`), only indeed, the reader part is missing. Would you mind reading it as such instead of a simple `/t`?

zobo:
Hi! Thanks for your comment. As far as I can tell from ECMA-376 there are two different `w:tab` elements. The one already defined in the code you refer to, I believe, is the one present in `w:pPr/w:tabs` . There is another `w:tab` element, defined in 17.3.3.32 chapter of `Ecma Office Open XML Part 1 - Fundamentals And Markup Language Reference.pdf`. This one doesn't have the same properties, however I couldn't find the relevant XML definition...
I noticed an example at the end of the chapter 11.3.9, page 50. It shows there that within one `w:r` there can be a combination of `w:t` and `w:tab` elements. I have changed the code to support such case.

If you thing a special element is needed anyway, I can try to figure out how to do that and define a different `Tab` class.

troosan:
indeed, you are right. You are referring to http://www.datypic.com/sc/ooxml/e-w_tab-2.html

PR #1319, opened 2018-03-22
Improve robustness of HTML parsing on invalid html

samimussbach:
### Description

When trying to create docx from invalid html markup docx creation crashes. I suppose it would be better if in this case all tags (aka formatting) is stripped but at least the text is kept and crash avoided as invalid markup might be produced by mysiwyg editors.
This PR addresses this. There is space for improvement e.g. in logging. Another glitch is that the strings are concatenated without space (see test).

Test testChartElements()  fails atm but it seems not to be related to my PR

Fixes # (issue)

### Checklist:

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -65,7 +65,11 @@ class Html
         // Load DOM
         $dom = new \DOMDocument();
         $dom->preserveWhiteSpace = $preserveWhiteSpace;
-        $dom->loadXML($html);
+        try {
+            $dom->loadXML($html);
+        } catch (\Exception $e) {
+            $dom->loadXML(strip_tags($html, '<body>'));
+        }
         self::$xpath = new \DOMXpath($dom);
         $node = $dom->getElementsByTagName('body');
 
--- a/tests/PhpWord/Shared/HtmlTest.php
+++ b/tests/PhpWord/Shared/HtmlTest.php
@@ -421,6 +421,21 @@ class HtmlTest extends \PHPUnit\Framework\TestCase
     }
 
     /**
+     * Tests parsing of invalid HTML
+     */
+    public function testParseInvalidHtml()
+    {
+        $phpWord = new \PhpOffice\PhpWord\PhpWord();
+        $section = $phpWord->addSection();
+        $html = '<p>This is some text with a paragraph</p><p>And another one that is not closed';
+        Html::addHtml($section, $html);
+
+        $doc = TestHelperDOCX::getDocument($phpWord, 'Word2007');
+
+        $this->assertEquals('This is some text with a paragraphAnd another one that is not closed', $doc->getElement('/w:document/w:body/w:p/w:r[1]/w:t')->nodeValue);
+    }
+
+    /**
      * Test parsing of img
      */
     public function testParseImage()


troosan:
I do not think it should be up to the library to "fix" the HTML, why not doing the try/catch in your code.
This way the client can treat the invalid HTML as he wants. Catching the exception in the library makes this impossible.

samimussbach:
I see your point but I think I do not follow you. The user of the library has no possibility to react reasonably when the exception is thrown at one of many items.
Say we have a list of 5 Paragraphs and the fourth one has invalid markup. At the moment an exception would be thrown and no docx would be created at all, even not with the other 4 paragraphs.
 
If it would be the case that an event would be thrown and the user of the library could handle the exception "inline" during paragraph parsing I definitely would be with you. But as it is implemented at the moment I would prefer to be able to generate an valid docx for the cost of loosing formatting.
Would it then be an option to make it configurable whether to be "robust" (or sloppy) or not?

troosan:
your solution is kind of all or nothing anyway no?

samimussbach:
I don't know whether I understand you correctly but if I do: No it isn't. Because I catch this specific Exception, the text is at least outputted, the treatment of any valid html remains the same. 
To be able to react It would of cause be way better to fire an Event in case of an Exception, get the modified payload of the Event (if the user modified it), try again to parse and if this fails again let the exception fire without catching it.
What do you think about that? Do you already have event hooks within phpword?

troosan:
Anyway, your solution does not work as `loadXml` does not throw an exception, it emits a warning instead. In order to be able to catch an exception you need to change the error handler first [see this post on php.net](http://www.php.net/manual/en/domdocument.loadxml.php#69295).

It works in a unit test because PHPUnit is doing exactly that and is throwing a PHPUnit_Framework_Error_Warning

samimussbach:
Indeed but strangely enough it works for me on a real project, not only in unittest. I‘ll dive in it.  
Is it worth working on this PR or would you say you would not want to change xml parsing anyway?

troosan:
For me the way to fix this is more like this:
```php
try {
    \PhpOffice\PhpWord\Shared\Html::addHtml($section, $html);
} catch (WhateverException $e) {
    $html = cleanupHtml($html);
    \PhpOffice\PhpWord\Shared\Html::addHtml($section, $html);
}
```
Maybe would could even add an operation in Html to add directly a DomDocument, like this
```php
public static function addHtmlFromDom(AbstractContainer $element, \DOMDocument $dom) {...}
```
This way it's up to the client to parse the HTML and pass it to PhpWord once cleaned up.
What do you think?

samimussbach:
Actually you are completely right. Somehow I did oversee that I am able to catch the exception thrown in addHtml and do the cleanup myself just like you suggested above.
I like the idea of introducing an addHtmlFromDom() as it is no big change and provides more flexibility if (and only if) needed.

samimussbach:
I understood, why I got an exception: Symfony was catching the warning and transforming it into an exception. Insofar this PR is not valid anyway, I'll close it. 
Should I open a new pull request with an custom errorhandler to be able to throw an exception like you suggested or would you?

troosan:
feel free so do it if you have the time :-)

PR #1343, opened 2018-04-11
Add support for table indent

Trainmaster:
### Description

This adds support for table indent (http://www.datypic.com/sc/ooxml/e-w_tblInd-1.html)

### Checklist:


Trainmaster:
@troosan I'd suggest to remove the `unit` in the table style options and also use `\PhpOffice\PhpWord\ComplexType\TblWidth` for the `width` option. Imho it makes no sense to use the same unit for both table style options.

troosan:
Indeed, I agree, it makes sense to use the TblWidth everywhere.
If you do the change, make sure you keep the getters/setters to keep backward compatibility.
Just mark them as deprecated so we do not forget to remove them in version 1.0

Trainmaster:
@troosan Alright. But I'll do that in a separate pull request.

PR #1354, opened 2018-04-23
(WIP) Fix cloneblock replaceblock

czosel:
### Description

Updated version of #541

Fixes # (issue)

### Checklist:


czosel:
Would be great if someone could take a first look before I'm making all the tests pass :smiley: 
@troosan

czosel:
Thanks for your feedback @troosan!

> wouldn't it be better to parse the tempDocumentMainPart to XML only once

Definitely - I took this shortcut for now just to avoid having to refactor even more :smile: But if we agree that this is the way to go forward, it's definitely worth it to make sure the parsing is only done once.

> The only problem with this refactoring is that it will probably make most of the currently open pull requests (add image, delete row, ...) very hard to merge.

You're right - but I guess in the long run it makes most sense to make sure that we're building on a solid foundation.

nguyenk:
First of all thank you so much for your work.

I'm very impatient to see your PR merged, because this might solve some of my problems in using phpword.

What is the current state of that PR ? Can I do something to make it go further ?

czosel:
Hi @nguyenk! Unfortunately, I don't think I'll find the time to finish this PR in the near future. While current state seems to work, it still needs to be optimized (parse to XML only once) and it lacks proper testing. It would be awesome if you could try to continue work on this - let me know if there is anything I could help with!

PR #1360, opened 2018-04-26
Drop GitHub pages, switch to coveralls

### LINKED ISSUE #1355: .git folder is huge
I got a timeout when I was trying to install PHPWord with composer. It seems that the `.git` folder is huge:
```bash
$ du -h --max-depth=1
2.1M	./src
1.9M	./tests
144K	./docs
1.2M	./samples
336M	./.git
```
Can someone tell how this happened, and if there would be a chance to do some history-rewriting to slim the `.git` folder down?

If you like, this could also be considered a composer issue (see https://github.com/composer/composer/issues/3449).

czosel:
Fixes #1355


czosel:
@troosan I just noticed that you disabled pushing the coverage after builds in the `develop` branch yesterday. Do you that goes far enough? Is the `gh-pages` branch still needed. I just confirmed locally that removing the `gh-pages` branch (and [cleaning up](https://stackoverflow.com/questions/34657031/how-do-you-remove-a-branch-and-all-of-its-history-locally-and-remotely) afterwards) reduces the size of `.git` to 11M.

czosel:
This change would probably also mean moving from scrutinizer to something like coveralls.

czosel:
@troosan Looks like coveralls Integration is working! Can you try adding it to the repo?

composer.json
```diff
@@ -61,7 +61,8 @@
         "php": "^5.3.3 || ^7.0",
         "ext-xml": "*",
         "zendframework/zend-escaper": "^2.2",
-        "phpoffice/common": "^0.2"
+        "phpoffice/common": "^0.2",
+        "php-coveralls/php-coveralls": "1.1.0"
```
troosan: shouldn't this be in require-dev instead?

troosan:
And there is first coveralls report!
@Progi1984 @czosel !

PR #1376, opened 2018-05-22
Add tracking changes for Heading or Title in case of docx reader 

moromete:
Add tracking changes for Heading or Title for docx reader
Also added tracking changes in case of Title for HTML writer

### DIFF
--- a/src/PhpWord/Writer/HTML/Element/Title.php
+++ b/src/PhpWord/Writer/HTML/Element/Title.php
@@ -26,6 +26,8 @@ use PhpOffice\PhpWord\Settings;
  */
 class Title extends AbstractElement
 {
+    use TrackChangeTrait;
+
     /**
      * Write heading
      *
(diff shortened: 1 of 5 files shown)


troosan:
nice @moromete, you know you can run the test/checks locally by running `composer run-script check --timeout=0`

Could you please check the code coverage report? There are some new lines not covered.
https://coveralls.io/builds/17125863

src/PhpWord/Writer/HTML/Element/Title.php
```diff
@@ -49,7 +50,68 @@ public function write()
...
+
+        return $content;
+    }
+
+    /**
+     * writes the track change opening tag
+     *
+     * @return string the HTML, an empty string if no track change information
+     */
+    private function writeTrackChangeOpening()
```
troosan: should avoid code duplication and move those 2 methods to the parent AbstractElement

PR #1380, opened 2018-05-23
Fix calls to the getEndingNotes method in the samples

WouterSioen:
### Description

This method call required both an array of writes and the filename, which was
missing in two method calls.

### Checklist:

---

I could not really check the last two items on the checklist since they are not really relevant.


WouterSioen:
Build failure seems unrelated to my changes, do I re-push my commit to trigger a new build?

PR #1383, opened 2018-05-23
Update documentation for Style::add*Style methods

hultberg:
### Description

Methods `addFontStyle`, `addParagraphStyle`, `addLinkStyle`, `addNumberingStyle`, `addTitleStyle`, and `setDefaultParagraphStyle` does allow `\PhpOffice\PhpWord\Style\AbstractStyle` to be passed as `Style::setStyleValues` checks for this on line 193.

### Checklist:

### DIFF
--- a/src/PhpWord/Style.php
+++ b/src/PhpWord/Style.php
@@ -95,7 +95,7 @@ class Style
      */
     public static function addTitleStyle($depth, $fontStyle, $paragraphStyle = null)
     {
-        if ($depth === null || $depth === '' || $depth === 0) {
+        if (empty($depth)) {
             $styleName = 'Title';
         } else {
             $styleName = "Heading_{$depth}";
--- a/src/PhpWord/Style.php
+++ b/src/PhpWord/Style.php
@@ -95,7 +95,7 @@ class Style
      */
     public static function addTitleStyle($depth, $fontStyle, $paragraphStyle = null)
     {
-        if ($depth == null) {
+        if ($depth === null) {
             $styleName = 'Title';
         } else {
             $styleName = "Heading_{$depth}";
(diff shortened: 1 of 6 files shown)


troosan:
Indeed, you are right. By fixing this you uncovered another issue. Can you check https://scrutinizer-ci.com/g/PHPOffice/PHPWord/inspections/80dcf145-3775-4219-a0f6-5fb9b4b0b59f/issues/files/src/PhpWord/Style.php?status=new&orderField=path&order=asc&honorSelectedPaths=0 and fix that as well?

hultberg:
@troosan I have changed the documentation of `Font::__constructor` now.

hultberg:
@troosan I have also changed to allow `null` on `$depth` on `Style::addTitleStyle`

src/PhpWord/Style.php
```diff
@@ -95,7 +95,7 @@ public static function addNumberingStyle($styleName, $styleValues)
      */
     public static function addTitleStyle($depth, $fontStyle, $paragraphStyle = null)
     {
-        if ($depth === null) {
+        if ($depth === null || $depth === '' || $depth === 0) {
```
hultberg: Could we just use `empty($depth)` here? it would catch all these cases.

troosan: indeed, that would definitely be cleaner

hultberg:
@troosan Looks like php 5.3 travis CV fails during composer install here as on develop branch.

troosan:
indeed, it seems finding the dependencies for (and with) php 5.3 is consuming a lot of memory (1.5Gb).
I'll try to restrict the ranges a bit for 5.3

PR #1430, opened 2018-07-16
allow to override TemplateProcessor#ensureUtf8Encoded

nicoder:
the method is `protected`,
but since it is called with `self` instead of `static`
it does not allow for subclasses to override it

(we have a special case of mostly ISO-8859-1 strings sometimes containing MS Word apostrophes,
and these appear as question marks in the document when `utf8_encode` is used,
but are displayed fine when we use `mb_convert_encoding($string, 'UTF-8', 'windows-1252')`)

An alternative would be to make the `ensureUtf8Encoded` non static,
since it only appears to be used in a non static method.

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -217,10 +217,10 @@ class TemplateProcessor
 
         if (is_array($replace)) {
             foreach ($replace as &$item) {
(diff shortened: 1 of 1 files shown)


troosan:
indeed, it would make sense to not make that method static. Being a java developer, it feels wrong to override a static method :-S

PR #1441, opened 2018-07-24
Adding table layout to the generated HTML

aarangara:
### Description

Adding table layout to the generated HTML if table element has layout style defined. This is useful when using creating PDF from PHPWord (e.g. using dompdf), otherwise the PDF does not contain any layout for table, and then it does not display any table columns which are exceeds the page width.

Fixes # (issue)

### Checklist:


troosan:
@aarangara I allowed myself to rewrite your solution a bit as the docx value for auto is autofit, which would not work with HTML (where it should be just auto).

PR #1468, opened 2018-09-13
Escape HTML entities from incoming HTML.

anotherjames:
### Description

HTML entities (e.g. `&lt;`) within text in HTML documents being read get decoded (e.g. into `<`), which means they can end up output wrong. In the case of `&lt;...&gt;`, the text between the two can go missing altogether in the output, as it is considered a HTML tag, rather than plain text.

Just the equivalent change was needed for text coming from XML/docx files, which was added in https://github.com/PHPOffice/PHPWord/commit/d8387c1abad2e0e44af3f936cad0e6d5746ad271 , I believe it's the same for text coming HTML.

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -267,7 +267,8 @@ class Html
         }
 
         if (is_callable(array($element, 'addText'))) {
-            $element->addText($node->nodeValue, $styles['font'], $styles['paragraph']);
+            $textContent = htmlspecialchars($node->nodeValue, ENT_QUOTES, 'UTF-8');
+            $element->addText($textContent, $styles['font'], $styles['paragraph']);
         }
     }
 


tjallingt:
Also running into this, would like to see a fix for this...

There doesn't seem to be an easy option (that actually works) to prevent DomDocument from decoding the html entities so fixing the problem like this seems to be the only working solution...

tjallingt:
@troosan Can you merge this?

troosan:
@anotherjames @tjallingt I'm afraid I do not agree with this change:
Adding the following `This &lt;text&gt; is not seen as an HTML tag, this &amp;lt; should be printed as HTML entity` should produce
`'This <text> is not seen as an HTML tag, this &lt; should be printed as HTML entity`
which it currently does.
Your change makes it impossible to print a `<` character for instance.

Or am I missing something? Can you provide an sample HTML you are trying to parse?

tjallingt:
Hey @troosan,

This example produces an broken docx for me:
```php
<?php

\PhpOffice\PhpWord\Shared\Html::addHtml($section, "Hello &amp; Bye");
\PhpOffice\PhpWord\Shared\Html::addHtml($section, "Hello & Bye");

$phpWord->save('test.docx');
```

It generates a word document containing the following **invalid** xml structure:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<w:document xmlns:ve="http://schemas.openxmlformats.org/markup-compatibility/2006" xmlns:o="urn:schemas-microsoft-com:office:office" xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships" xmlns:m="http://schemas.openxmlformats.org/officeDocument/2006/math" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:wp="http://schemas.openxmlformats.org/drawingml/2006/wordprocessingDrawing" xmlns:w10="urn:schemas-microsoft-com:office:word" xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main" xmlns:wne="http://schemas.microsoft.com/office/word/2006/wordml">
  <w:body>
    <w:p>
      <w:pPr/>
      <w:r>
        <w:rPr/>
        <w:t xml:space="preserve">Hello & Bye</w:t>
      </w:r>
    </w:p>
    <w:p>
      <w:pPr/>
      <w:r>
        <w:rPr/>
        <w:t xml:space="preserve">Hello & Bye</w:t>
      </w:r>
    </w:p>
    <w:sectPr>
      <w:pgSz w:orient="portrait" w:w="11905.511811023622" w:h="16837.79527559055"/>
      <w:pgMar w:top="1440" w:right="1440" w:bottom="1440" w:left="1440" w:header="720" w:footer="720" w:gutter="0"/>
      <w:cols w:num="1" w:space="720"/>
    </w:sectPr>
  </w:body>
</w:document>
```
You can see the whole document here: [test.docx](https://github.com/PHPOffice/PHPWord/files/2940332/test.docx)

The `&` character is not allowed anywhere in XML and should be replaced with its Entity counterpart `&amp;`. The same thing happens for `<` and `>`. Somehow it doesn't happen for `'` and `"` which AFAIK are also not allowed.

The solution is that any html strings supplied to addHtml should first be html decoded and then re-encoded so our example would go through the following steps:

```js
addHtml("Hello &amp; Bye")  // -> "Hello & Bye" -> "Hello &amp; Bye" 
addHtml("Hello & Bye")      // -> "Hello & Bye" -> "Hello &amp; Bye"
```

Since browsers have a tendency to just accept invalid XML structures the current HTML output works when viewed in a browser: 
```html
<!DOCTYPE html>
<!-- Generated by PHPWord -->
<html>
  <head>
    <meta charset="UTF-8" />
    <title>PHPWord</title>
    <style>
      * {
        font-family: Arial;
        font-size: 10pt;
      }
      a.NoteRef {
        text-decoration: none;
      }
      hr {
        height: 1px;
        padding: 0;
        margin: 1em 0;
        border: 0;
        border-top: 1px solid #ccc;
      }
      table {
        border: 1px solid black;
        border-spacing: 0px;
        width: 100%;
      }
      td {
        border: 1px solid black;
      }
    </style>
  </head>
  <body>
    <p style="margin-top: 0; margin-bottom: 0;">Hello & Bye</p>
    <p style="margin-top: 0; margin-bottom: 0;">Hello & Bye</p>
  </body>
</html>
```

But here `&amp;` would work and technically be more correct.

tjallingt:
Ah I forgot to apologise for not responding earlier 😊
I hope we can re-open this or another pull request to get this issue fixed.
Note that I didn't test whether this PR fixes the issue but it seems related

troosan:
@tjallingt All you need to do is first call
```php
\PhpOffice\PhpWord\Settings\Settings::setOutputEscapingEnabled(true);
```

PR #1470, opened 2018-09-17
Recognise style differences within headings.

anotherjames:
### Description

Inline style differences within headings can currently be missed when reading a docx file. For example, when a heading style is bold, but part of the heading text has been un-bolded. This change ensures that styling that is due to the heading style, is also set on individual Text elements within a TextRun, so they can be differentiated from any that are explicitly overridden.

Previously, all the Text elements within a TextRun that gets set on a Title element, just get the default Font element. That meant, for example, that `isBold()` returned false, whereas heading styles are often in bold.

### DIFF
--- a/src/PhpWord/Reader/Word2007/AbstractPart.php
+++ b/src/PhpWord/Reader/Word2007/AbstractPart.php
@@ -23,6 +23,8 @@ use PhpOffice\PhpWord\Element\AbstractContainer;
 use PhpOffice\PhpWord\Element\TextRun;
 use PhpOffice\PhpWord\Element\TrackChange;
 use PhpOffice\PhpWord\PhpWord;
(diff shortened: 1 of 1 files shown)


troosan:
Please add unit test and fix syntax

troosan:
@anotherjames As far as I could test your solution only partially works.
I modified the document opened by Sample_11_ReadWord2007.php, making part of the heading not bold. But after parsing it the whole title is still bold. Making part of the title italic works.
So your solution only works when "adding" style, not when "removing" style.

troosan:
@anotherjames I did some more testing. Could you check the sample_11, you will see that
- Heading 1 is declared as being bold
- Part of the "PHP Word" is un-bolded, and is Italic in addition

PR #1493, opened 2018-10-25
RTF writer: Round getPageSizeW and getPageSizeH to avoid decimals

Patrick64:
When output as RTF the decimal places of pagesize are coming out at the start of the document. See https://stackoverflow.com/questions/48391201/numbers-show-at-beginning-of-phpword-document

This only changes it for RTF writer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,19 @@ php:
     - 7.0
     - 7.1
     - 7.2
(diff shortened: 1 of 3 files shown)


Patrick64:
Related to https://github.com/PHPOffice/PHPWord/pull/1459

PR #1496, opened 2018-11-01
Add number in parameter for cloneBlock function

snapter:
Add squencial number at the end of each parameter inside a cloning block

### Description

Fixes # (issue)

### Checklist:


snapter:
> 
> 
> @snapter could you explain why this is needed? What does it fix?

@troosan when cloning the block that is contained multiple parameters, the parameter should be cloned and replace with the cloned number.

troosan:
@snapter OK, I see what you're doing.
I'll close this PR as it is the same as #459 and #817, which was already merged.

Issue #1500, opened 2018-11-07
ampersand (&) in HTML corrupts word document file


silverbackdan:
This is:

- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpword)

### Expected Behavior

Correct generation of a word file

### Current Behavior

A word file is generated but cannot be opened because it is corrupted.

### Failure Information

When adding HTML, if there is an ampersand (just an & character) the output document is corrupt. It is required to turn the ampersand into html encoded &amp;

### How to Reproduce
```php
$section = $this->phpWord->addSection();
Html::addHtml($section, '&', false, false);
```

### Context

* PHP version: 7.2.7
* PHPWord version: 0.15

silverbackdan:
I've had a quick look and the html normalization seems a little too simplistic as well.

Instead of
`$html = str_replace('&', '&amp;', $html);` how about something like `$html = preg_replace('/(&)(?![0-9a-z]+;)/i', '&amp;', $html);` This way you will not catch html characters already defined. TBH I'm still not really sure why the ampersand causes this issue. Does it need to be escaped further down the line when inserting into the word document format?

troosan:
The following is normal it fails, as this is not valid HTML.
```php
Html::addHtml($section, '&', false, false);
```
The following works fine
```php
Html::addHtml($section, '&amp;', false, false);
```
As for the bug with the addText function, indeed, this should be escaped when writing to XML, but I don't think the Element/Text class is the place to do this. You will break the RTF writer. This should be escaped in the Word2007 writer instead.

silverbackdan:
When the dom element is created in php, &amp; is converted to just `&` so there's no getting around it I don't think. There is a normalization attempt made already that I spotted but it is reverted when getting the html from the php dom element.

I am not sure what the final document formats look like, I've not researched that before so am not sure what breaking changes there are between output formats. Could you link me to the function that you'd like me to move the fix to? I'm happy to do that if you'd like - I'm also not precious about the fix should you want to move it to where you think it is most appropriate.

Thanks for your time on this.

jeromeWeissmann:
Hi, got the same error here.
I've tried the fallowing:
`Html::addHtml($this->section, "Hello &amp; Bye");`
Or 
`Html::addHtml($this->section, "Hello & Bye");`

docx file is corrupt in both cases.

silverbackdan:
[...]

PR #1501, opened 2018-11-07
Fix ampersand corrupting document

silverbackdan:
see: https://github.com/PHPOffice/PHPWord/issues/1500

### Description

This will replace any ampersand that is not already followed by a semi-colon (making it an HTML character) with `&amp;`. This is because without converting it, the document cannot be opened.

Fixes # (issue)
https://github.com/PHPOffice/PHPWord/issues/1500

### Checklist:

- [ ] I have update the documentation to describe the changes **n/a**

### DIFF
--- a/src/PhpWord/Element/Text.php
+++ b/src/PhpWord/Element/Text.php
@@ -56,6 +56,8 @@ class Text extends AbstractElement
      */
     public function __construct($text = null, $fontStyle = null, $paragraphStyle = null)
     {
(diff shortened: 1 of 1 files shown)


silverbackdan:
FYI: PHP 5.3 test fails due to php running out of memory in travis.

troosan:
@silverbackdan wouldn't it be better to do this replacement when generating the document instead of when creating the Text element? What will happen when generating not a docx but an rtf document?

margori:
Patch applied, bug not solved.
Suggest include test.

silverbackdan:
right ok, thanks for the additional info - I appreciate everyone's time in replying on this chain. Enjoy your week.

Issue #1529, opened 2018-12-07
addListItemRun with paragraph syles now working


kreciszd:
This is:

- [ ] a bug report

### Expected Behavior

Should render correct list with paragraph styles, for rgular ListItems it is working

### Current Behavior

![zrzut ekranu z 2018-12-07 14-49-40](https://user-images.githubusercontent.com/30601886/49651432-7057a880-fa2f-11e8-90b0-3bb980d0c702.png)

### Failure Information

List items has the same paragraph styles so they should render the same, but for the list with text runs that does not do it
### How to Reproduce

Please provide a code sample that reproduces the issue.

```php
$predefinedMultilevelStyle = array('listType' => \PhpOffice\PhpWord\Style\ListItem::TYPE_NUMBER_NESTED);
$paragraphStyleName = 'P-Style';
$phpWord->addParagraphStyle($paragraphStyleName, array('spaceBefore' => 400, 'alignment' => 'end'));

$section->addText('Multilevel predefined list. Test');
$section->addListItem('List Item 1', 0, $fontStyleName, $predefinedMultilevelStyle, $paragraphStyleName);
$section->addListItem('List Item 2', 0, $fontStyleName, $predefinedMultilevelStyle, $paragraphStyleName);
$section->addListItem('List Item 3', 1, $fontStyleName, $predefinedMultilevelStyle, $paragraphStyleName);
$section->addListItem('List Item 4', 1, $fontStyleName, $predefinedMultilevelStyle, $paragraphStyleName);
$section->addListItem('List Item 5', 2, $fontStyleName, $predefinedMultilevelStyle, $paragraphStyleName);
$section->addListItem('List Item 6', 1, $fontStyleName, $predefinedMultilevelStyle, $paragraphStyleName);
$section->addListItem('List Item 7', 0, $fontStyleName, $predefinedMultilevelStyle, $paragraphStyleName);
$section->addTextBreak(2);

$section->addText('List with inline formatting. Test');
$listItemRun = $section->addListItemRun(0, $predefinedMultilevelStyle, $paragraphStyleName);
$listItemRun->addText('List item 1');
$listItemRun->addText(' in bold', array('bold' => true));
$listItemRun = $section->addListItemRun(1, $predefinedMultilevelStyle, $paragraphStyleName);
$listItemRun->addText('List item 2');
$listItemRun->addText(' in italic', array('italic' => true));
$footnote = $listItemRun->addFootnote();
[...]

PR #1628, opened 2019-05-09
No nested w:pPr elements in ListItemRun.

waltertamboer:
### Description

This commit fixes issue #1529

This commit prevents nested w:pPr elements when using a ListItemRun with
a paragraph style. The different between a ListItem and a ListItem run
is that the setWithoutPPR method is called on the ParagraphStyleWriter
(PhpOffice\PhpWord\Writer\Word2007\Style\Paragraph).

According to the specs it's not allowed to have nested w:pPr elements.
See http://www.datypic.com/sc/ooxml/e-w_pPr-2.html

### Checklist:

### DIFF
--- a/src/PhpWord/Writer/Word2007/Element/ListItemRun.php
+++ b/src/PhpWord/Writer/Word2007/Element/ListItemRun.php
@@ -17,6 +17,7 @@
 
 namespace PhpOffice\PhpWord\Writer\Word2007\Element;
 
+use PhpOffice\PhpWord\Element\ListItemRun as ListItemRunElement;
 use PhpOffice\PhpWord\Writer\Word2007\Style\Paragraph as ParagraphStyleWriter;
 
 /**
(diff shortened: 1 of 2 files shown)


src/PhpWord/Writer/Word2007/Element/ListItemRun.php
```diff
@@ -31,34 +32,56 @@ class ListItemRun extends AbstractElement
...
+    private function writeParagraphProperties(ListItemRunElement $element)
+    {
+        $xmlWriter = $this->getXmlWriter();
         $xmlWriter->startElement('w:pPr');
-        $paragraphStyle = $element->getParagraphStyle();
-        $styleWriter = new ParagraphStyleWriter($xmlWriter, $paragraphStyle);
+
+        $styleWriter = new ParagraphStyleWriter($xmlWriter, $element->getParagraphStyle());
         $styleWriter->setIsInline(true);
+        $styleWriter->setWithoutPPR(true);
```
waltertamboer: The magic.

waltertamboer:
I've also created smaller methods to make the code a bit more easy to follow.

PR #1646, opened 2019-07-03
Change private to protected to be able extending class Html

SpinyMan:


Dwarfex:
@troosan What is needed to merge this?

PR #1669, opened 2019-07-18
Check lengths, colors, styles, etc automatically

0b10011:
### Description

**I will be committing and rebasing often until this gets to a more usable state. I just wanted to open a draft pull request in the meantime to avoid anyone else doing the same work.**

PHPWord uses a mixture of strings (constants and otherwise), integers, floats, and magic to track widths, heights, colors, styles, etc. Unfortunately, it's currently impossible to check the correct units are used, and colors and styles are often not validated.

This is an attempt to move PHPWord to a more OOP codebase, where input is verified, lengths can be provided using any unit, and conversion for import/export happens automatically.

For example, currently, you can add a row to a table with `$table->addRow(900)`. However, there's no unit provided with the height and the documentation for `addRow` doesn't list which unit should be used. While the unit could be listed in the documentation fairly easily, that doesn't help to avoid user errors. With this new approach, adding a row would use `$table->addRow(Length::from("twip", 900))`. It is now easy to see which unit is used from the calling code, and the receiving code no longer has to worry about which unit was provided, as it can use `$height->toInt("twip")`, guarnateeing the correct unit is used.

Fixes # (issue)

While I haven't dug through the issues to find any existing github issues this fixes, I have fixed several bugs along the way that would not have been noticed otherwise (except by checking by hand). And I imagine it's fixing a lot more issues that have gone unnoticed for quite some time.

### Checklist:


0b10011:
Tests are passing locally, but there's a good chance some of them will fail in CI. I'll get back to fix any that pop up, and will be double checking coverage to make sure new stuff is covered properly.

In the meantime, feel free to review the code, approach, etc. I'll only be adding tests and fixing issues that pop up, so it should be stable enough to review. I did one round of documentation updates already, but it probably needs another one. Let me know if you find anything, and I'll see what I can do!

0b10011:
Merging #1661 broke this. I'll get a fixed version up tomorrow. (Tried using GitHub's in-browser conflict resolution, but it resulted in an empty commit.)

0b10011:
Rebased, fixed, and squashed. All checks passing.

troosan:
@0b10011 Impressive work. I'm just a bit worried about the non-backward compatibility of the changes.
This one will make it very difficult to migrate to a next version. It would be nice to be able to still pass sizes/colors/... the "old" way, and transform internally.

0b10011:
> @0b10011 Impressive work.

Thanks! I didn't quite realize how large the change was going to be until I was already well into it, but I'm pretty happy with how it turned out.

>  I'm just a bit worried about the non-backward compatibility of the changes.
> This one will make it very difficult to migrate to a next version. It would be nice to be able to still pass sizes/colors/... the "old" way, and transform internally.

Agreed! Like I said, I didn't realize how large of a project was until I'd already done a lot of changes, so it isn't friendly to upgrading as-is.

Unfortunately, we'd lose the benefit of enforced types this way. I thought about manually checking types, but that would add a *lot* of boilerplate for the sole purpose of upgrading (which I know and believe is important to support somehow). I didn't do this for two reasons:

1. The tests and code is already written, so it'd be a lot of work to go back in and add support for strings.
2. This work would eventually be reverted, but it would have to be reverted by hand most likely if any meaningful amount of time passed.

What I *think* would work even better would be:

1. Merge this PR with enforced types for the **next version**
2. Merge a separate PR into the **old/current version** that coerces `Length`, `Color`, etc into the expected strings.

This will still be a lot of work, but then it will only need to be done once. That is, it won't need to be reverted later.

It's been a little bit since I submitted this, but I think there were some minor API changes along the way too. That should be able to be reflected in the current version as well.

Anyway, I'm more than happy to do this work as well. I just want to make sure the approach is good, the class signatures won't change, etc. I can even put together a proof of concept if you'd like to show what I mean. Just let me know!

PowerKiKi:
This looks like a super interesting thing PR. Would you be willing to rebase on develop ?

My plan would be to merge this, and release as a major version, and do nothing to help migration (!). The community might come up with a [Rector](https://getrector.org/) rule for that...

0b10011:
> This looks like a super interesting thing PR. Would you be willing to rebase on develop ?

I unfortunately have covid right, and it's got me pretty worn down.

After I recover, it could be a fun project though! I'll set a reminder for myself in a week or so, and see if it's something I can take care of. (I'm no longer working on the project I was hoping to use this for, though, so no promises.)

I remember there being a bunch of other things that could use strict typing to greatly improve the library, too, but I can't remember exactly what. I'll try to note those things if I'm able to work on it.

PowerKiKi:
Nothing matter more than your health. Take all the time you need.

And if you find out that you're not so much interested in this anymore, that's totally fine too. Just let us know, so someone else might be picking it up later on...

PR #1674, opened 2019-07-24
Remove urldecode to handle filenames with spaces

MegaphoneJon:
### Description

Converting an HTML file to Word fails when the HTML contains an `` tag whose `<src>` is a) a URL, and b) contains a space.  

I'm not sure if there's a reliable way to test this, since it requires loading a file from a URL; the bug isn't present when a space exists in a file path.  However, I confirm that this passes all existing tests.

Fixes # (none)

### Checklist:

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -686,7 +686,6 @@ class Html
                 fclose($ifp);
             }
         }
-        $src = urldecode($src);
 
         if (!is_file($src)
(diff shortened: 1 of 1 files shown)


MegaphoneJon:
Relevant quote from the [PHP docs for file_get_contents](https://www.php.net/manual/en/function.file-get-contents.php):
> If you're opening a URI with special characters, such as spaces, you need to encode the URI with urlencode().

PR #1707, opened 2019-08-28
Fix apt-get crash in Travis CI for PHP 5.3

### LINKED ISSUE #1706: Travis CI pipeline is broken for PHP 5.3
### Describe the Bug

When running the Travis CI job for PHP 5.3 under Ubuntu Precise, it crashes during the before_install step, when running `apt-get update`.

It is because Ubuntu Precise (12.04) is EOL for a long time and the MongoDB repo is not available anymore. However Ubuntu 12.04 is the only one where PHP 5.3 can still run.

BTW is it still worth supporting PHP 5.3? Also, since Precise is EOL it is likely that more and more breakages will occur, would it be worth switching to Xenial (default in Travis), at least for the PHP versions that it supports (5.6+)?

If it makes past this step, it then crashes at `composer install --prefer-source` because of exceeded memory usage, 2G is no longer enough it seems.

### Steps to Reproduce

- Open a PR
- Look at Travis CI log for PHP 5.3

### Expected Behavior

The build should complete.

### Current Behavior

The build crashes during the operation `sudo apt-get update -qq` with the error message "E: Failed to fetch http://downloads-distro.mongodb.org/repo/debian-sysvinit/dists/dist/InRelease  Clearsigned file isn't valid, got 'NOSPLIT' (does the network require authentication?)"

Once fixed, it then crashes during `composer install --prefer-source` with the message "PHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 32 bytes)"

### Context

- PHP Version: 5.3
- PHPWord Version: any

mdupont:
### Description

Fixes the crash when calling `apt-get update` in PHP 5.3 environment in Travis CI. Since the culprit is a 3rd party repo for MongoDB and that we don't need MongoDB, I just remove the repo definition so `apt-get` doesn't try to connect to it.

Fixes #1706

### Checklist:

- [X] I have updated the documentation to describe the changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
         - php: 7.0
           env: COVERAGE=1
         - php: 5.3
-          env: COMPOSER_MEMORY_LIMIT=2G
+          env: COMPOSER_MEMORY_LIMIT=3G
         - php: 7.3
           env: DEPENDENCIES="--ignore-platform-reqs"
     exclude:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ env:
 
 before_install:
     ## Packages
+    - sudo rm -f /etc/apt/sources.list.d/mongodb.list  # Makes apt crash on Precise, and we don't need MongoDB
     - sudo apt-get update -qq
     - sudo apt-get install -y graphviz
 


mdupont:
Travis build worked as expected, but is still reported as failed because of phpmd warning that were already existing in the code in the develop branch, nothing related to this PR.

mdupont:
If requirements get bumped to PHP >= 7.0 (and we can benefit from the enhanced type declarations), then indeed there won't be any issue anymore :-)

stephanvierkant:
Wait, does this library still supports PHP versions from the middle ages?

PR #1712, opened 2019-09-02
Added return type

fmasa:
### Description

There was return type currently missing. Also property type was wrong as `$parent` can be null (i.e. for `Section`)

### Checklist:


fmasa:
I just noticed that maybe type can be refined to `AbstractContainer|null` as only instances of `AbstractContainer` can have child elements.

PR #1755, opened 2019-11-17
Improvements in RTF writer

oleibman:
1. Codes meant to be in hex are specified in decimal.
   Consequently characters which don't need escaping are escaped.
2. Special handling (prepend backslash) needed for {, }, and \.
   RTF docs generated with those characters cannot be opened in Word.
3. Tab character needs to be escaped as \tab.
   RTF docs drop these characters.
While running test suite, found that Writer/RTF/ElementTest was coded
only for Unix line endings, and fails on Windows. Changed so that it
would work on either.

### Description

Fixes # (issue)

### Checklist:

### DIFF
--- a/src/PhpWord/Escaper/Rtf.php
+++ b/src/PhpWord/Escaper/Rtf.php
@@ -26,8 +26,14 @@ class Rtf extends AbstractEscaper
 {
     protected function escapeAsciiCharacter($code)
     {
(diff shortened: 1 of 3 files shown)


oleibman:
No documentation change is required. I see that some people have been able to convert that checkbox to the equivalent of "not applicable", but I don't see how to do that, so I've just checked it for now.

PR #1774, opened 2019-12-06
Add font style for Field elements

oleibman:
### Description

Fixes # (issue)

### Checklist:

### DIFF
--- a/docs/elements.rst
+++ b/docs/elements.rst
@@ -403,7 +403,9 @@ Currently the following fields are supported:
 
 .. code-block:: php
 
-    $section->addField($fieldType, [$properties], [$options], [$fieldText])
+    $section->addField($fieldType, [$properties], [$options], [$fieldText], [$fontStyle])
+
+- ``$fontStyle``. See :ref:`font-style`.
 
 See ``\PhpOffice\PhpWord\Element\Field`` for list of properties and options available for each field type.
 Options which are not specifically defined can be added. Those must start with a ``\``.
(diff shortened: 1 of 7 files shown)


oleibman:
RTF Changes

1. Converter is currently expecting colors as strings of hex digits,
   but PhpWord allows specification of colors by named constant, so
   result is random when one of those is used. This change handles
   all the named colors.
2. Table needs \pard at end; formatting may be wrong without it.
3. RTF writer will no longer ignore paragraph style for TextRun.
4. RTF writer will no longer ignore paragraph and font style for Title.
5. Add support for RTF headers and footers.
6. Add support for right-to-left in font.
7. Add support for PageBreakBefore and LineHeight for paragraphs.
8. Add support for PageNumberingStart for sections.

There are test cases for all of these changes.

I apologize - I do not know why my commit message, which is now shown immediately above,
was lost.

oleibman:
I'm not sure if I need to do something about "resolve conflicts". Tests weren't working because fixes were required to AbstractWebServerEmbeddedTest.php, for which there was a pull request for a change which had not yet been merged, so I borrowed the code from that change. Now that it has been merged, it no longer needs to be part of this change, but I'm not sure how to disentangle it. Please let me know what I should be doing, if anything.

oleibman:
Still having trouble with the commit message. Trying to re-enter what it should have been with this
message..

Add font style for Field elements

Add the ability to use a Font Style with Field elements, and have Word2007 writer use that style.

Also, Word seems to require that w:rtl be specified inline rather than as part of a named style.
This change satisfies that requirement by generating inline w:rtl tag when it is part of a named style.

Test cases are added for both changes.
Changed documentation to note that Field can now have Font Style.

PR #1780, opened 2019-12-13
Limit parameter not working for TemplateProcessor setImageValue

FireworksWebsites:
Fixes and issue where the limit parameter is not correctly limiting the maximum amount of tags to replace on setImageValue


troosan:
already merged #1967

PR #1799, opened 2020-01-06
Improved Word2007 Reader to read images when pictures is between <wp:anchor>

begnini:
### Description

Word2007 Reader didn't read images where pictures was between <wp:anchor>, accepction only with <wp:inline>. 

The OpenOffice XML docs says:

> The XML for placement of the picture (e.g., whether it is inline [**<wp:inline>**] or anchored [**<wp:anchor>**]) is within the wordprocessing drawingML namespace: xmlns:wp="http://schemas.openxmlformats.org/drawingml/2006/wordprocessingDrawing".

Improved tests to be capable of read images from docs generated by ```AbstractTestReader::getDocumentFromString```

### Checklist:


begnini:
This solves the same problem pointed by PR #1794. The difference between the PRs is the tests I added.

Issue #1800, opened 2020-01-07
CloneBlock regexp does not replace the block if line-drop


JuzSer:
### Describe the Bug

When you have multiple Blocks in multiple Rows, some of blocks cannot be replaced
Only the last block was replaced. 
This because some word file will drop-line after `xml` tag, and the regular expression engine in different versions cannot find the matched block.

### Steps to Reproduce

See the image below, only blocks #6 has been replaced. 
- **TemplateProcessor** 
- The template has a table with clone rows. Each row has different dynamic data. 
- The blocks name in each row will be `block#1`, `block#2`,... 

<img width="314" alt="screenshot" src="https://user-images.githubusercontent.com/5735071/71869492-d1771180-3144-11ea-8556-7de419406245.png">

```php
<?php
...
$this->processor->cloneRow('client', $amount);
for (...) {
    $suffix = '#' . ($i + 1);
    // block_year.
    $this->processor->cloneBlock('block_year' . $suffix);
}
```

### Expected Behavior

Replaced all Blocks 

### Current Behavior

Only blocks in the last row was replaced

### Context

Please fill in your environment information:

- PHP Version: 7.3.9
- PHPWord Version: 0.17.0

### Fix & Pull Request:
`TemplateProcessor.php`
Line: 740

Change the regexp pattern to: 
```php
'/(.*((?s)<w:p\b(?:(?!<w:p\b).)*?\${' . $blockname . '}<\/w:.*?p>))(.*)((?s)<w:p\b(?:(?!<w:p\b).)*?\${\/' . $blockname . '}<\/w:.*?p>)/is',
```

Describe: 
- Do not find for xml.
- Start find the `<w:p` tag that closest to the blockname.
- This `<w:p` does not contain any other `<w:p` inside it.

After fix: 
<img width="298" alt="fixed" src="https://user-images.githubusercontent.com/5735071/71870169-84486f00-3147-11ea-926f-3be5ba3c25bf.png">

Hope this helps.

JuzSer:
Update the regexp to avoid catastrophic backtracking

`'/(.*((?s)<w:p\b(?:(?!<w:p\b).)*?\${' . $blockname . '}<\/w:.*?p>))(.*)((?s)<w:p\b(?:(?!<w:p\b).)[^$]*?\${\/' . $blockname . '}<\/w:.*?p>)/is',`

cawolf:
[...]

PR #1801, opened 2020-01-07
Fix: CloneBlock regexp for different regexp engine with xml line-drop

juzser:
### Description
Reference Issue:   #1800 
When you have multiple Blocks in multiple Rows, some of blocks cannot be replaced
Only the last block was replaced.
This because some word file will drop-line after `xml` tag, and the regular expression engine in different versions cannot find the matched block.

Fixes # (issue)
`TemplateProcessor.php`
Line: 740

Change the regexp pattern to: 
```php
'/(.*((?s)<w:p\b(?:(?!<w:p\b).)*?\${' . $blockname . '}<\/w:.*?p>))(.*)((?s)<w:p\b(?:(?!<w:p\b).)*?\${\/' . $blockname . '}<\/w:.*?p>)/is',
```
Describe: 
- Do not find for xml.
- Start find the `<w:p` tag that closest to the blockname.
- This `<w:p` does not contain any other `<w:p` inside it.

### Checklist:
All Done

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -737,7 +737,7 @@ class TemplateProcessor
         $xmlBlock = null;
         $matches = array();
         preg_match(
-            '/(<\?xml.*)(<w:p\b.*>\${' . $blockname . '}<\/w:.*?p>)(.*)(<w:p\b.*\${\/' . $blockname . '}<\/w:.*?p>)/is',
+            '/(.*((?s)<w:p\b(?:(?!<w:p\b).)*?\${' . $blockname . '}<\/w:.*?p>))(.*)((?s)<w:p\b(?:(?!<w:p\b).)*?\${\/' . $blockname . '}<\/w:.*?p>)/is',
             $this->tempDocumentMainPart,
             $matches
         );


JuzSer:
Update PR: https://github.com/PHPOffice/PHPWord/pull/1809

Herz3h:
I encountered an I issue where my blocks were not cloned (basically preg_match didn't match anything). I changed regex to what's used here, and it was fixed. Can we merge this ?

JuzSer:
> I encountered an I issue where my blocks were not cloned (basically preg_match didn't match anything). I changed regex to what's used here, and it was fixed. Can we merge this ?

@Herz3h Feel free to use it. Or you can use my fork repo here: https://github.com/JuzSer/PHPWord

Herz3h:
Would prefer to have it merged, so I don't forget later on that I'm using your repo instead of the official one in my composer ;)

xu-li:
In my case, it works if [pcre.jit](https://www.php.net/manual/en/pcre.configuration.php#ini.pcre.jit) is enabled on php 7.3.6.  And it will also work if [pcre.backtrack-limit](https://www.php.net/manual/en/pcre.configuration.php#ini.pcre.backtrack-limit) is big enough when _pcre.jit._ is disabled.

Those might not be good solutions, but could be served as temporary solution before the official fix.

JuzSer:
> In my case, it works if [pcre.jit](https://www.php.net/manual/en/pcre.configuration.php#ini.pcre.jit) is enabled on php 7.3.6. And it will also work if [pcre.backtrack-limit](https://www.php.net/manual/en/pcre.configuration.php#ini.pcre.backtrack-limit) is big enough when _pcre.jit._ is disabled.
> 
> Those might not be good solutions, but could be served as temporary solution before the official fix.

@xu-li Yes, I see that problem before, long time ago I posted an update PR here: https://github.com/PHPOffice/PHPWord/pull/1809
This will solve the catastrophic backtracking. You can try this.

xu-li:
Thanks, @JuzSer . Yes, I saw your PR, tried that and it worked in my particular case. 

I didn't use your code because:
- We use phpword in production.
- Regular expression is a beast to me and I am not familiar with the details of cloneBlock. I couldn't evaluate if your regular expression will work for all the use cases.

It's just my personal preference. I want to play on the safe side. It has nothing to do with your code.

JuzSer:
Yes, of course. 
I'm just waiting for the package's owner to check this issue.. But seem likes he has no longer maintenance this package, maybe.

troosan:
merged #1809

PR #1816, opened 2020-02-04
Add center / right / left image position in template img set

Serizao:
Add possibility to define position of picture in array of properties. exemple :

```php
$templateProcessor->setImageValue('picture',array("path" => "exemple.jpg" , "align" => "center"));
```


4unkur:
@Serizao sorry to bother you, but do you have any ideas of how we can add `wrappingStyle = behind` functionality to `setImageValue()` method like it's in `addImage()` method of phpword?  

I tried to play with setComplexValue and use textRun with addImage - but image is shown as blank in the output file. There are several issues related to that.

Krato:
Is this going to be merged?

PR #1886, opened 2020-06-17
Add user defined node mappings

HughbertD:
### Description

At [GatherContent](https://gathercontent.com) we're currently prototyping with PHPWord to power some of our content exports. 

We want to leverage the `addHtml` functionality as this is the easiest content type we have to hand, however, the number of HTML tags that are implemented in this function are limited. 

Rather than trying to implement all the required tag parsing functionality; a simpler route for us would be to allow custom parsing mappings be created by the library user. 

This also solves a future problem for us, where we will want to create highly customised mappings for non-standard HTML tags specific to our domain, and we hope is a welcome addition for existing PHPWord users. 

Fixes # (issue)

### Checklist:

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -35,6 +35,7 @@ class Html
     protected static $listIndex = 0;
     protected static $xpath;
     protected static $options;
+    protected static $userDefinedNodeMappings = array();
 
     /**
      * Add HTML parts.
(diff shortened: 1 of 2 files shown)


rodrigoq:
Hello, this pull request is something I need too, I was wondering what is needed to merge? I can help solving the conflicts, but I don't understand exactly what's the problem. If I add this changes to the latest commit (if it's needed) and add some tests and docs would they be merged? Or is there some other problem to address (wrong api, wrong code standard, etc)?

PR #1902, opened 2020-07-11
Enhancements to addHTML parser

lubosdz:
### Description

Currently importing HTML via `PhpWord\Shared\Html::addHtml()` has couple of bugs and misses support for some at least basic HTML tags & attributes. This PR attempts to fix these issues without BC break (at least no existing unit test break):

### Fixed legacy issues:

* fix [passing](https://github.com/PHPOffice/PHPWord/blob/develop/src/PhpWord/Shared/Html.php#L364) `width` property into cells, allowing also to control colum widths inside tables. Following would currently not work `style="width: 35%"` - because width is passed as a wrong argument. Once fixed, it will also allow to control columns width inside tables. Since all widths are currently ignored (not recognized) they are set to `width = "auto"`.
* fix unit conversion for margins/spacing - instead of converting to [points](https://github.com/PHPOffice/PHPWord/blob/develop/src/PhpWord/Shared/Html.php#L588), it should convert to `twips`. Currently passing 10px results into 7.5 twips therefore value is seemingly ignored, unless users sets some absurd value like `width: 600px`. 

### New enhancements:

* support width also as an attribute e.g. `<td width="25%">`
* support cellspacing for tables e.g. `<table cellspacing="3">`
* support all `border-*` variants e.g. `border-bottom: 1px #DDDDDD solid;` (top, right, left, bottom) - useful for tables & cells
* support bgColor for rows, cells, e.g. `<tr bgColor="#DDDDDD">`
* support horiz. rule `<hr />` with limited custom styling
* support both inline vertical-align and valign for cells e.g. `<th valign="bottom">` or `style="vertical-align: middle;"`
* support start & numbering type in ordered lists e.g. `<ol type="A" start="3">`

### Checklist:

### More enhancements possible (with BC break)

A two more fixes should be fixed to make current method `addHtml()` more usable and compatible:

* fix `cell property inheritance` - currently all elements always inherit all properties from parent node. This however should not apply to cells. Cells should not inherit most of parent properties e.g. borders, margins etc. because it produces unexpected output. In spite of BC break this should be fixed ASAP. Current behaviour is quite incompatible with legacy HTML output. Moreover, inheriting widths from parent nodes causes unpredictable results in table layouts.
* add support for Headings `<h1> .. <h6>` out of the box. Current implementation relies that user will somewhere define his own heading style `Heading1`, `Heading2` etc., which adds extra code boilerplate for developers and also requires to study how to do that (increases learning curve). Since H1 .. H6 are very basic elements in any HTML, these should work out of the box.

Unfortunatelly, both improvements above will cause BC break, and therefore have been excluded from this PR in order to make space for discussion.

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -447,7 +447,31 @@ class Html
         } else {
             $data['listdepth'] = 0;
             $styles['list'] = 'listStyle_' . self::$listIndex++;
-            $element->getPhpWord()->addNumberingStyle($styles['list'], self::getListStyle($isOrderedList));
+            $style = $element->getPhpWord()->addNumberingStyle($styles['list'], self::getListStyle($isOrderedList));
+
+            // extract attributes start & type e.g. <ol type="A" start="3">
+            $start = 0;
+            $type = '';
+            foreach ($node->attributes as $attribute) {
+                switch ($attribute->name) {
+                    case 'start':
+                        $start = (int) $attribute->value;
+                        break;
+                    case 'type':
+                        $type = $attribute->value;
+                        break;
+                }
+            }
+
+            $levels = $style->getLevels();
+            /** @var \PhpOffice\PhpWord\Style\NumberingLevel */
+            $level = $levels[0];
+            if($start > 0){
+                $level->setStart($start);
+            }
+            if($type && !!($type = self::mapListType($type))){
+                $level->setFormat($type);
+            }
         }
         if ($node->parentNode->nodeName === 'li') {
             return $element->getParent();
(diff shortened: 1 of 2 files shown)


lubosdz:
I checked errors - not sure what could I fix here: 
* travis looks like environment config issue
* weird error on line 114 `A parse error occurred: Syntax error, unexpected EOF on line 114 at column 0` looks like false issue, since it's comment line. I ran tests on PHP 7.0.33 and PHP 7.4.1 without issues locally.

lubosdz:
There is 1 failing job - not sure what to do about it, does not seem related. Log:
~~~
Loading composer repositories with package information
Updating dependencies (including require-dev)
PHP Fatal error:  Allowed memory size of 1610612736 bytes exhausted (tried to allocate 72 bytes) in phar:///home/travis/.phpenv/versions/5.6.40/bin/composer/src/Composer/DependencyResolver/Rule2Literals.php on line 48
Fatal error: Allowed memory size of 1610612736 bytes exhausted (tried to allocate 72 bytes) in phar:///home/travis/.phpenv/versions/5.6.40/bin/composer/src/Composer/DependencyResolver/Rule2Literals.php on line 48
~~~

lubosdz:
Is this repo dead? No feedback since 3 weeks ago ... ?

devfake:
What's up with this PR? This adds some good features :|

lubosdz:
If you need a quick enhancement, you may want to use composer repo `https://github.com/lubosdz/PHPWord` - which holds these unmerged changes. But there is no guarantee it will not change. Or make clon for yourself.

This repo looks truly like abandoned, even though not officially. Probably time & money issue, like with most open sources ..

lubosdz:
@troosan 
Thank you for merging. 
Do you think it makes sense to make another PR with more serious BC break - as mentioned above? (support for headers H1 .. H6, fix  for table cells inheritance).

troosan:
@lubosdz feel free to create another PR with those changes. I won't be merging them in this release though, it's been waiting for a long time :-)

PR #1916, opened 2020-08-13
Update composer.json

wandrewchan:
remove abandoned zend-escaper, replace with laminas

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ php:
 matrix:
     include:
         - php: 5.6
-          dist: trusty
+          env: COMPOSER_MEMORY_LIMIT=4G
         - php: 7.0
           env: COVERAGE=1
         - php: 7.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,11 @@ php:
 matrix:
     include:
         - php: 5.6
-          env: COMPOSER_MEMORY_LIMIT=4G
+          env: COMPOSER_MEMORY_LIMIT=-1
         - php: 7.0
-          env: COVERAGE=1
+          env: COVERAGE=1 COMPOSER_MEMORY_LIMIT=-1
         - php: 7.3
-          env: DEPENDENCIES="--ignore-platform-reqs"
+          env: DEPENDENCIES="--ignore-platform-reqs" COMPOSER_MEMORY_LIMIT=-1
     exclude:
         - php: 5.3
         - php: 5.4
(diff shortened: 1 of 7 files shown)


stephanvierkant:
You also changedf some other files. Can you explain those changes?

internalsystemerror:
the latest version of laminas/laminas-escaper requires PHP >= 5.6

stephanvierkant:
> the latest version of laminas/laminas-escaper requires PHP >= 5.6

Of course it does. PHP 5.6 is very very old.

stephanvierkant:
> the latest version of laminas/laminas-escaper requires PHP >= 5.6

What's the point?

internalsystemerror:
U asked what the other changes were for. As far as I can see, those changes are removing the support for PHP < 5.6. That's the only point here.

stephanvierkant:
I mean files like `src/PhpWord/Escaper/AbstractEscaper.php`

liborm85:
Why is upgraded minimal php version? This is BC break. You can use `"laminas/laminas-escaper": "^2.2",`.

And PR contains other irrelevant changes in AbstractEscaper.php.
PR is wrong.

stephanvierkant:
Yes, in theory this might be a BC break. But it has been upgraded to a PHP version that hasn't got support for years now. I don't consider removing support for EOL-versions a BC-break.

wandrewchan:
Actually I was trying to replace with laminas as what @tacman, but it was failed. And I found that zend-escaper has deprecated, but it still works. So this merge is discard and closed.

liborm85:
I'm adding compatibility with PHP 7.4 and PHP 8.0 in PR https://github.com/PHPOffice/PHPWord/pull/1946, was necessary fix travis and scrutinizer. So also migrate to Laminas, commit: https://github.com/PHPOffice/PHPWord/pull/1946/commits/f2516b08b146da9b2be66326bb36d6f54fe56781

stephanvierkant:
That would be (at least) the third attempt to replace zend by laminas. And again, that PR contains other changes as well.

liborm85:
Without next changes it is not possible to test whether it works in travis.

tacman:
My PR was a one-line fix, and it gets rid of the warning that shows up every time.  Was there a problem with that PR?  Yes, at the moment the deprecated version still works, but there's no reason to keep it.

PHP 5 is dead, dead, dead.  Anyone using it should upgrade.  That's a different PR, of course.

PR #1936, opened 2020-09-22
Add null check when setComplexValue is not found

YannikFirre:
### Description

Fix when setComplexValue is not found in the template.


src/PhpWord/TemplateProcessor.php
```diff
@@ -274,6 +274,11 @@ public function setComplexValue($search, \PhpOffice\PhpWord\Element\AbstractElem
         $elementWriter->write();
 
         $where = $this->findContainingXmlBlockForMacro($search, 'w:r');
+
+        if($where === false) {
```
troosan, line 278: please fix missing space after if (see travis check)

YannikFirre: It's done :)

PR #1950, opened 2020-10-21
Limit of setImageValue

### LINKED ISSUE #1825: Limit of setImageValue not working
### Describe the Bug

I use the template processor, to edit an existing Word file. When I want to set an image, the defined limit is not being considered. If the limit in setImageValue() is set to 1, all occurences are replaced and not only one. However, the limit when using setValue() works fine.

### Steps to Reproduce

```php
<?php

$wordDoc = new TemplateProcessor($pathToDocument);
$wordDoc->cloneBlock('CLONEBLOCK', $numberOfElements);

foreach ($elements as $element) {
	$wordDoc->setValue('element_name', $element->getName), 1);
	$wordDoc->setImageValue('element_picture', $pathToPicture, 1);
}
```

### Expected Behavior

The image of an element should be only set for one occurence of 'element_picture'.

### Current Behavior

The image of the first element is set for all occurences of 'element_picture'.

### Context

- PHP Version: 7.4
- PHPWord Version: 0.17.0 via composer

ponsifiax:
### Description

When you try to setImage with limit, the processor apply correctly the limit instead replace all tags.

Fixes #1825

### Checklist:

- [ x] I have run `composer run-script check --timeout=0` and no errors were reported
- [x ] The new code is covered by unit tests (check build/coverage for coverage report)

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -585,6 +585,9 @@ class TemplateProcessor
                     return ($partVar == $searchString) || preg_match('/^' . preg_quote($searchString) . ':/', $partVar);
                 });
 
+                // Exclude multiple same variable name
+                $varsToReplace = array_unique($varsToReplace);
+
                 foreach ($varsToReplace as $varNameWithArgs) {
                     $varInlineArgs = $this->getImageArgs($varNameWithArgs);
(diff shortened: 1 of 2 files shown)


troosan:
I think #1967 is a better fix for this. It actually takes into account the "limit" parameter.
I like the unit tests though ;-)

ponsifiax:
I think too, but maybe keep the unit test to confirm the use case?

PR #1965, opened 2020-11-09
feat: Update addHtml to handle style inheritance

Julien1138:
### Description

The aim is to get the output closer to the source html.

I noticed the issue in the doc file generated from HTML, in the table, a border was applied to the text inside the cells. Which was not right because no border was applied to the span. I understood that [CSS inheritance of style properties](https://developer.mozilla.org/en-US/docs/Web/CSS/inheritance) was not taken into account by the parser so I decided to implement it.

### Checklist:

- [ ] I have updated the documentation to describe the changes **(Nothing to update)**


Julien1138:
There seems to be an issue with the .travis.yml file... Some kind of typo maybe but I'm no expert.

Also, this pr relates to #1902.

PR #1971, opened 2020-11-20
Fix TemplateProcessor::fixBrokenMacros to handle whitespaces

### LINKED ISSUE #590: PHPWord randomly deletes whitespaces
PHPWord seems to randomly delete whitespaces from my docx Documents. In the templates, the spaces are there. But when I render them in PHPWord, the new Document has some whitespaces missing. Anyone else have this problem?

chapa:
### Description

This is a proposal to fix #590 in a cleaner way than a global regex on the whole document, which might change parts of the document that must not change.

*Edit: see https://github.com/PHPOffice/PHPWord/pull/1971#issuecomment-970481810*

`TemplateProcessor::fixBrokenMacros()` is a method that strip all tags between `${`, the name of the variable and `}` (let's call them the 3 parts). If there is a `<w:t xml:space="preserve">` in those tags it means that there is a trailing space at the end of the content of the tag, because there is not supposed to be any space between the 3 parts, so the whole must be inside of it in order to keep this trailing space.

Examples:
```xml
<w:r><w:t>before ${</w:t></w:r><w:r><w:t xml:space="preserve">variable} </w:t></w:r><w:r><w:t>after</w:t></w:r>
                                                                       |
                                                              the trailing space

<w:r><w:t>before ${</w:t></w:r><w:r><w:t>variable</w:t></w:r><w:r><w:t xml:space="preserve">} </w:t></w:r><w:r><w:t>after</w:t></w:r>
                                                                                             |
                                                                                    the trailing space
```

I think it's safe to move the opening tag just before the variable, but one thing I'm not sure about because I don't know Open XML very well, is whether it's safe to prepend the variable by `</w:t></w:r><w:r><w:t xml:space="preserve">`. I think it's ok if a `<w:t>` is necessarily in a `<w:r>` and there cannot be multiple `<w:t>` in a `<w:r>` (otherwise we could just remove the `</w:r><w:r>` part).

Expected result of the above examples:
```xml
<w:r><w:t>before </w:t></w:r><w:r><w:t xml:space="preserve">${variable} </w:t></w:r><w:r><w:t>after</w:t></w:r>
```

### Checklist:

=> it's a bug fix

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -1070,7 +1070,11 @@ class TemplateProcessor
         return preg_replace_callback(
             '/\\' . $brokenMacroOpeningChars . '(?:\\' . $endMacroOpeningChars . '|[^{$]*\>\{)[^' . $macroClosingChars . '$]*\}/U',
             function ($match) {
-                return strip_tags($match[0]);
+                preg_match_all('/<.+?\s*\/?\s*>/si', $match[0], $tags);
+                $tags = implode('', $tags[0]);
+                $tags = str_replace('<w:t>', '<w:t xml:space="preserve">', $tags);
+
+                return strip_tags($match[0]) . $tags;
             },
             $documentPart
         );
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -1063,14 +1063,24 @@ class TemplateProcessor
      */
     protected function fixBrokenMacros($documentPart)
     {
-        $brokenMacroOpeningChars = substr(self::$macroOpeningChars, 0, 1);
-        $endMacroOpeningChars = substr(self::$macroOpeningChars, 1);
-        $macroClosingChars = self::$macroClosingChars;
-
         return preg_replace_callback(
-            '/\\' . $brokenMacroOpeningChars . '(?:\\' . $endMacroOpeningChars . '|[^{$]*\>\{)[^' . $macroClosingChars . '$]*\}/U',
+            sprintf(
+                '/%s.+%s/U',
+                implode('', array_map(
+                    function (string $char) {
+                        return preg_quote($char) . '(?:<[^>]+>)*';
+                    },
+                    str_split(self::$macroOpeningChars)
+                )),
+                implode('', array_map(
+                    function (string $char) {
+                        return '(?:<[^>]+>)*' . preg_quote($char);
+                    },
+                    str_split(self::$macroClosingChars)
+                ))
+            ),
             function ($match) {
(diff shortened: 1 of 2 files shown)


4rthem:
Thank you for the fix!
Could it be merged?

chapa:
Actually I found a case where my fix isn't working: when there is styling tags between the `$` and the `}` of a single variable. Because all tags are stripped we lose styling information.

After several tests, I came to the conclusion that the best way to fix those broken macros without breaking anything else was to move all existing tags inside the variable right after it, for example:
```xml
<w:r><w:t>${</w:t></w:r><w:r><w:t>variable1</w:t></w:r><w:r><w:t>}</w:t></w:r>
<!--        └─────────1──────────┘         └─────────2──────────┘          -->
```
should be transformed to:
```xml
<w:r><w:t>${variable1}</w:t></w:r><w:r><w:t></w:t></w:r><w:r><w:t></w:t></w:r>
<!--                  └─────────1──────────┘└─────────2──────────┘         -->
```
That way we just reassemble all parts of the macro together.

But we also need to add `xml:space="preserve"` to all `<w:t>` we move, to handle this situation:
```xml
<w:r><w:t>${variable1</w:t></w:r><w:r><w:t>} ${variable2</w:t></w:r><w:r><w:t>}</w:t></w:r>
```
that would be transformed to:
```xml
<w:r><w:t>${variable1}</w:t></w:r><w:r><w:t xml:space="preserve"> ${variable2}</w:t></w:r><w:r><w:t xml:space="preserve"></w:t></w:r>
<!--                                        └──────────────────┘ │                                  └──────────────────┘          -->
<!--                                            we need this     └─ to preserve this space        (this one does not hurt)        -->
```

I updated the fix to handle this

will2877:
Will there be any movement on this PR in the near future?
I'm experiencing isses with missing whitespaces all over my documents.

chapa:
Gently ping @troosan

@will2877 until this is merged you can use [`cweagans/composer-patches`](https://packagist.org/packages/cweagans/composer-patches) with the patch at https://github.com/PHPOffice/PHPWord/commit/f955a818014be369a50371da2445dda1c6aaeaf0.patch

chapa:
I rebased the branch to be up to date with `master`, this needed some adaptations to handle custom macro chars that wasn't overridable at the time the PR was made.

@Progi1984 can you take a look ?

chapa:
Hey @Progi1984, I saw some activity on this repo, let me know if you want me to rebase this PR and adapt it to the latest version if needed.

But you can also let me know if you want me to close it, as it seems to be ghosted since I opened it 4 years ago.

chapa:
Thanks @adq-talbot, for what it worth I'm using this fix in production since I wrote it and I haven't noticed any problems so far.

Issue #294, opened 2014-06-26
Support for paragraph with borders


CauanCabral:
There's a way to define a paragraph border (size, color)?
If not, how can I help include that feature? I'd read Element/Table and Style/Table as base and tried changed Style/Paragraph to extends Border but didn't work.

PR #1986, opened 2020-12-19
add/setting page element border style.

emnabs:
In the [issue#294](https://github.com/PHPOffice/PHPWord/issues/294) problem, it is found that a piece of code is applied, but the disadvantage is that the border style setting is not added when setting the border, so the code of adding paragraph style is supplemented. And added a list of static constants for border styles.

### DIFF
--- a/src/PhpWord/Style/Border.php
+++ b/src/PhpWord/Style/Border.php
@@ -23,41 +23,6 @@ namespace PhpOffice\PhpWord\Style;
 class Border extends AbstractStyle
 {
     /**
-     * Border Style Val.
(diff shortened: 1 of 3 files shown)


src/PhpWord/Style/Border.php
```diff
@@ -22,6 +22,41 @@
  */
 class Border extends AbstractStyle
 {
+    /**
```
troosan, line 25: Please move to a PHPWord/SimpleType/Border
source is http://www.datypic.com/sc/ooxml/t-w_ST_Border.html

PR #1987, opened 2020-12-20
Please set the scope of the TemplateProcessor::replaceXmlBlock `public` instead of `protected`

cyrilrouyer:
### Description

The goal is to be able to directly replace a macro with a content that is more than a simple AbstractElement (for example : text followed by a table).

setComplexValue and setComplexBlock are not suitable for this kind of needs: the project I'm workin' on can't just have multiple macros instead of just one, because the body of the macro will depends on the context.

For the moment I just created a class that inherits TemplateProcessor, in order to be able to access the replaceXmlBlock method, but it would be more cleaner if the method was directly public.


troosan:
please change the target of this PR, it should merge into develop, not master

troosan:
PR on wrong branch.

PR #1992, opened 2020-12-23
Fix image stroke in libreoffice 7.x

Adizbek:
### Description
Recently I've updated to libreoffice 7.x and noticed weird bug, by default ms word and libreoffice 6.x and below show no line, but starting from libreoffice 7.x border appeared around images which was inserted using `setImageValue`

MsWord
image

Libreoffice 7.x
image


thisnugroho:
Thank you~, i was trying to figuring out this problem x_x

PR #2006, opened 2021-01-20
Fix cloneBlock issue

lexdewilligen:
### Description

The regex function in cloneBlock may be too complicated or result in too many matches. Therefore, it can occur that the function results in no matches and therefore doesn't replace any blocks. I've solved this issue by making the regex more precise (by adding an extra group) and this solves the issue. It could probably also be fixed by increasing one of [these]
(https://www.php.net/manual/en/pcre.configuration.php) PHP configurations. This however is not what should be expected from users of this package.

Might fix many cloneBlock and replaceBlock issues
https://github.com/PHPOffice/PHPWord/issues?q=cloneBlock
https://github.com/PHPOffice/PHPWord/issues?q=replaceBlock

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -781,14 +781,14 @@ class TemplateProcessor
     {
         $matches = array();
         preg_match(
-            '/(<\?xml.*)(<w:p.*>\${' . $blockname . '}<\/w:.*?p>)(.*)(<w:p.*\${\/' . $blockname . '}<\/w:.*?p>)/is',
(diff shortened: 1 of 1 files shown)


troosan:
@lexdewilligen is this issue still present in 0.18.0?
might it have been fixed by https://github.com/PHPOffice/PHPWord/pull/1809 ?

andcarpi:
Fixed my problem. After cloning a row that had a cloneblock inside, it only clonned the last 7 blocks... 
i.e.: 
```
clonerow('test', 10)
for(i=1;i<11;i++) {
   cloneblock('im_inside_row#'.i,5)
}
```
would only clone rows 8,9 and 10

thePanz:
@troosan I can confirm that the issue is still present with the latest release 0.18.2

This PR fixes the issue. Thank you @lexdewilligen !

PR #1946, opened 2020-10-17
Compatibility with PHP 7.4, PHP 8.0 and migrate to Laminas Escaper

liborm85:
### Description

- compatibility with PHP 7.4
- compatibility with PHP 8.0 (for full support require update in 3rd libraries)
- travis-ci: fixed execution on PHP 5.x
- fixed scrutinizer, allow zip extension
- migrate from abandoned `Zend\Escaper` to `Laminas Escaper`

For full compatibility with php 8.0 require PR https://github.com/PHPOffice/Common/pull/34

Fixes https://github.com/PHPOffice/PHPWord/issues/1944
Fixes https://github.com/PHPOffice/PHPWord/issues/1874
Fixes https://github.com/PHPOffice/PHPWord/issues/1797

### Checklist:

- [x] I have run `composer run-script check --timeout=0` and no errors were reported
- [x] The new code is covered by unit tests (check build/coverage for coverage report)
- [x] I have updated the documentation to describe the changes

PR #2032, opened 2021-03-04
Fix BC break in #1946. This package does not replace laminas/laminas-zendframework-bridge

### LINKED ISSUE #2028: Replacing laminas/laminas-zendframework-bridge without providing its functionality breaks packages
### Describe the Bug
In f2516b08b146da9b2be6 (#1946) was introduced that this package replaces `laminas/laminas-zendframework-bridge`. I do not really understand why this package should be replaced by phpword, which is especially weird as newly required `laminas/laminas-escaper` itself requires `laminas/laminas-zendframework-bridge`

In my opinion #1946 breaks setups where other libraries need `laminas/laminas-zendframework-bridge`. I am a bit hesitating, as no one else seems to have this problem. Perhaps anybody could help me to understand, what happens here?

### Steps to Reproduce

Install phpword 0.18 in a setup with a package requiring the bridge. Then call anything.

### Expected Behavior

Nothing should break

### Current Behavior

A fatal error occurs `Attempted to load class "ClassGenerator" from namespace "Zend\Code\Generator".
Did you forget a "use" statement for e.g. "Laminas\Code\Generator\ClassGenerator"`

### Context

- PHP Version: 7.3
- PHPWord Version:  0.18

mussbach:
### Description

This commit fixes the BC break introduced by #1946 where it was stated
that phpword replaces package laminas/laminas-zendframework-bridge which
is not the case. Package laminas/laminas-zendframework-bridge only was
replaced from phpword.

Fixes #2028 (issue)

### Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,9 +75,6 @@
         "mpdf/mpdf": "5.7.4 || 6.* || 7.* || 8.*",
         "php-coveralls/php-coveralls": "1.1.0 || ^2.0"
     },
-    "replace": {
-        "laminas/laminas-zendframework-bridge": "*"
-    },
     "suggest": {
         "ext-zip": "Allows writing OOXML and ODF",
         "ext-gd2": "Allows adding images",


PowerKiKi:
Thank you for this. It's released as [0.18.1](https://github.com/PHPOffice/PHPWord/releases/tag/0.18.1)

PR #2041, opened 2021-03-17
Update Html parser to accept line-height:normal

joelgo:
For acept text in line-height how "line-height:normal", etc.

### Description

Fixes # (issue)

### Checklist:

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -661,16 +661,9 @@ class Html
                     break;
                 case 'line-height':
                     $matches = array();
-                    if (preg_match('/([a-z]+)/', $cValue, $matches)) {
-                        //$cvalue text and not number
-                        if ($cValue == 'normal') {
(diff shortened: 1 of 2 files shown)


src/PhpWord/Shared/Html.php
```diff
@@ -661,7 +661,21 @@ protected static function parseStyle($attribute, $styles)
                     break;
                 case 'line-height':
                     $matches = array();
-                    if (preg_match('/([0-9]+\.?[0-9]*[a-z]+)/', $cValue, $matches)) {
+                    if (preg_match('/([a-z]+)/', $cValue, $matches)) {
+                        //$cvalue text and not number
+                        if($cValue == 'normal') {
```
troosan, line 666: what else than 'normal' can it be? And why 1.12 for one and 1.13 for the other?

troosan (changes requested):
please fix issues

PR #2043, opened 2021-03-21
check if macro is inside table

GustavoViniciusdeMorais:
### Check if macro is inside table

It was added a function to verify if the macro used to clone a table row inside the document is inside a table to avoid the bug 'Can not find the start position of the row to clone.'

It gives to the user of the library a way to avoid the throw of a new exception and continue processing the document without an error caused by a mistake from his client.

Fixes # (improvement)

### Checklist:


troosan:
please fix style issues

PR #2052, opened 2021-04-10
Update chart, don't write 'c:overlap' if grouping is 'clustered'

dfsd534:
### Description
when bar or column grouping by clustered have c:overlap,the bar will be overlap
Fixes # (issue)


troosan:
please fix issues

PR #2119, opened 2021-07-27
TemplateProcessor can accept a file contents instead of the file path

LemarinelNet:
### Description

In many projects, Devs are now using Cloud storage (such as AWS S3). 
So, having a word template in a S3 storage would force the dev to save it to a custom/temporary path and then give the path to the TemplateProcessor constructor.

I submit this PR so the constructor can ALSO accept the file contents. 
If the given parameter is a path to a valid file, then the constructor works as before. Else, we suppose that the parameter contains the file contents, and then we save it to the tempDocumentFilename, and use it as we did before.

### Checklist:

### DIFF
--- a/docs/templates-processing.rst
+++ b/docs/templates-processing.rst
@@ -6,6 +6,7 @@ Templates processing
 You can create an OOXML document template with included search-patterns (macros) which can be replaced by any value you wish. Only single-line values can be replaced.
 Macros are defined like this: ``${search-pattern}``.
 To load a template file, create a new instance of the TemplateProcessor.
+The constructor can accept the path to the Word Template, or the contents of the file (ie. if you had to load it from AWS S3, http, FTP, etc...)
 
 .. code-block:: php
 
--- a/samples/Sample_42_TemplateWithFileContents.php
+++ b/samples/Sample_42_TemplateWithFileContents.php
@@ -1,0 +1,23 @@
+<?php
+include_once 'Sample_Header.php';
+
+// Template processor instance creation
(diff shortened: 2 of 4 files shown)


waltertamboer (changes requested):
I tried to help by reviewing the PR.

I do understand the motivation for making this change, yet I would recommend a different approach. Instead of providing the path to a file or the content of a file, it would be better to approach this from a more OOP kind of way:

```
interface TemplateProcessorFileLoader
{
    /**
     * Loads the file and returns a stream to the file that should be processed.
     *
     * @return string Returns a stream to the file content.
     */
    public function load(): resource;
}
```

```
class TemplateProcessor
{
    // ... snip ...
    public function __construct(TemplateProcessorFileLoader $loader)
    {
        $this->tempDocumentFilename = $this->loadFile($loader);
        
        // snip
    }

    private function loadFile(TemplateProcessorFileLoader $loader): string
    {
        $tempDocumentFilename = tempnam(Settings::getTempDir(), 'PhpWord');

        if (false === $tempDocumentFilename) {
            throw new CreateTemporaryFileException(); // @codeCoverageIgnore
        }

        // Something like this or figure out whatever approach works best...
        $handle = fopen($tempDocumentFilename, 'wb');
        // ... error handling ...
        stream_copy_to_stream($loader->load(), $handle);
        fclose($handle);

        return $tempDocumentFilename;
    }
```

Nevertheless it might be a good idea to merge this PR since it does add more flexibility and create an issue to refactor this for a next major version.

Issue #2121, opened 2021-07-28
Internal Link feature


damienfa:
### Is your feature request related to a problem? Please describe.

In Word2007, it is possible to make a link to an existing heading in the document. 
If I load a .docx file with PhpWord (throught a Reader) , and I rewrite it immediately (Writer), these internal references to titles are lost because PhpWord does not manage them. 

### Describe the solution you'd like

It's a pity, because everything can be done simply in the Reader : detect the "anchors" as internal links, and make sure that they point to the same reference (w:name) on the title's Bookmark. 

### Describe alternatives you've considered

I think that it is directly the Title object that should know the "w:name" of its internal bookmark. But the Title object does not talk about bookmark...
I see that in the Writer of Word2007, a bookmark is created arbitrarily with a "w:name=_ToC" whereas if Title knew the ref of its bookmark, it could create it in a more intelligent way and improve its management of internal links.
 
### Additional context

To test with the attached file, you will see that the internal links are lost.

```
$docPath = "/app/doc-with-internal-links.docx";
$phpWord = \PhpOffice\PhpWord\IOFactory::load($docPath);
$objWriter = \PhpOffice\PhpWord\IOFactory::createWriter($phpWord, 'Word2007');
$objWriter->save('/app/doc-with-NO-internal-link.docx');
```
Attached file : [doc-with-internal-links.docx](https://github.com/PHPOffice/PHPWord/files/6889823/doc-with-internal-links.docx)

damienfa:
@begnini , @troosan , is there a reason why you force "Heading_1" as a title style name, instead of using the original style name used in the docx document ? 


https://github.com/PHPOffice/PHPWord/blob/ab7d82274eb72c8380b86bfd0b59f931da53783e/tests/PhpWord/Reader/Word2007/StyleTest.php#L193

In french, the original style name for a title is "Titre1", in spanish is "Titulo1" etc.
The title style are lost in the French Word application when you re-open a file loaded->written by PhpWord. 
I have some easy solutions for that but I need to understand why the "Heading_4" format is forced ?

PR #2122, opened 2021-07-28
Detect Word Internal Links and add title bookmark

damienfa:
### Description

At the moment, PhpWord is not able to read an internal link to an existing heading. 
This is because the anchor is not detected in the reader. Here I create a new (internal) "Link" when this attribute is found.
Moreover, the Headings of the Word document usually contain a "bookmark" (useful for TOC or internal link) but the reader has not created it so that the name (reference of the bookmark) is lost.

Fixes # (issue)

https://github.com/PHPOffice/PHPWord/issues/2121

### Checklist:

### DIFF
--- a/src/PhpWord/Reader/Word2007/AbstractPart.php
+++ b/src/PhpWord/Reader/Word2007/AbstractPart.php
@@ -160,6 +160,15 @@ abstract class AbstractPart
                     $this->readRun($xmlReader, $node, $textContent, $docPart, $paragraphStyle);
                 }
             }
+            
+            // Every heading is generaly associated to a bookmark. 
+            // Get the name of the current title bookmark and add a similar bookmark before the title 
+            $bookmarkTitleNodes = $xmlReader->getElements('w:bookmarkStart', $domNode);
+            if ($bookmarkTitleNodes->length) {
+                $bookmarkName = $bookmarkTitleNodes->item(0)->getAttribute('w:name');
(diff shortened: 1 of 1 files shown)


src/PhpWord/Reader/Word2007/AbstractPart.php
```diff
@@ -160,6 +160,15 @@ protected function readParagraph(XMLReader $xmlReader, \DOMElement $domNode, $pa
                     $this->readRun($xmlReader, $node, $textContent, $docPart, $paragraphStyle);
                 }
             }
+
+            // Every heading is generaly associated to a bookmark.
+            // Get the name of the current title bookmark and add a similar bookmark before the title
+            $bookmarkTitleNodes = $xmlReader->getElements('w:bookmarkStart', $domNode);
+            if ($bookmarkTitleNodes->length) {
```
waltertamboer, line 167: You should only use booleans in an if condition: `if ($bookmarkTitleNodes->length !== 0) {`

waltertamboer: There is a problem with the `getElements` method. The DocBlock of `getElements` says it returns a `DOMNodeList` but it actually can also return `false` or an `array`. You should check for that here as well (or let the getElements method throw an exception if it returns false).

waltertamboer (commented):
Nice change, I tried to help by reviewing your PR. I added some small nitpicking comments but overall I think it's good to be merged.

PR #2123, opened 2021-07-28
Custom Word title style name (ex: "Titre 1") taken into account

damienfa:
### Issue / Problem

At the moment, when the heading of document are not named "Heading 1", "Heading 2" in the Word document, then PhpWord does not consider them as heading. When the document is loaded (through a Reader), there is no "Title" element.

It's a common problem for French, Spanish, or other user.  
Indeed, in these countries, the local version of Microsoft Word defines default titles named "Title 1" (in France), "Titulo 1" (in Spanish) ... and it is impossible to replace them into Word ... 

This should not be a problem for PhpWord which has the indication that it is a "heading" via the attribute "w:name" but which should be consistent with the style identifier. See below how a french docx document is generated by Word : 

```
<w:style w:type="paragraph" w:styleId="Titre1">
        <w:link w:val="titre1Char"/>
        <w:name w:val="heading 1"/>
        <w:basedOn w:val="Normal"/>
```

### Description

The solution I propose here is, when creating a heading style, to keep the original name of the style provided in the docx (w:styleId), but to force the "w:name" to a value of the format "heading 1".

This impacts: 
- the Reader: which must now read the specific "w:styleId" from the docx and provide this value to the Style;
- Style.php : which must know how to create a style with the custom name provided ;
- the Writer : which must know how to reuse the custom name of the heading style, but force the "w:name" to indicate that it is a heading;

### Checklist:

### DIFF
--- a/src/PhpWord/Reader/Word2007/Styles.php
+++ b/src/PhpWord/Reader/Word2007/Styles.php
@@ -65,8 +65,9 @@ class Styles extends AbstractPart
             foreach ($nodes as $node) {
                 $type = $xmlReader->getAttribute('w:type', $node);
                 $name = $xmlReader->getAttribute('w:val', $node, 'w:name');
+                $styleId = $xmlReader->getAttribute('w:styleId', $node);
                 if (is_null($name)) {
-                    $name = $xmlReader->getAttribute('w:styleId', $node);
+                    $name = $styleId;
                 }
                 $headingMatches = array();
                 preg_match('/Heading\s*(\d)/i', $name, $headingMatches);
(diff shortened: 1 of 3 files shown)


damienfa:
@begnini , @troosan , is there a reason why you force "Heading_1" as a title style name, instead of using the original style name used in the docx document ? 
I thought it was for "ease", but I see unit tests about it, so it's starting to puzzle me 🤔

https://github.com/PHPOffice/PHPWord/blob/ab7d82274eb72c8380b86bfd0b59f931da53783e/tests/PhpWord/Reader/Word2007/StyleTest.php#L212

In french, the original style name for a heading is "Titre1", in spanish is "Titulo1" etc.
The title styles are lost in the French Word application when you re-open a file loaded->written by PhpWord. 

I have some easy solutions for that but I need to understand why the "Heading_1" format is forced ?

Also check this Pull Request : https://github.com/PHPOffice/PHPWord/pull/2121

PR #2193, opened 2022-02-08
#2187 Addresses deprecation on PHP 8.1 in ZipArchi

### LINKED ISSUE #2187: Deprecation on PHP 8.1 in ZipArchive
```
Deprecated: ZipArchive::open(): Passing null to parameter #2 ($flags) of type int is deprecated in /var/itrans/one/app/Vendor/phpoffice/phpword/src/PhpWord/Shared/ZipArchive.php on line 136
```

The current method signature of `ZipArchive::open()` is `public function open($filename, $flags = null)`.

If no argument is passed as the second argument here, `null` is passed as the second argument on line 136:

```
            $zip = new \ZipArchive();
            $result = $zip->open($this->filename, $flags); // <-- line 136
```

Passing `0` achieves the same behaviour as passing `null`. Therefore, I believe the appropriate fix here would be:

```
$zip = new \ZipArchive();

if ($flags === null) {
 $flags = 0;
}

$result = $zip->open($this->filename, $flags);
```

This would eliminate the deprecation and not break any BC within the library.

I'm happy to submit a PR if required.

garethellis36:
### Description

This addresses the deprecation in PHP 8.1 noted in issue #2187.

The fix does not break backwards compatibility because existing code can still pass no second argument (or pass explicitly pass `null`) and existing behaviour will not be affected.

Fixes #2187 

### Checklist:

I didn't add any new tests. There is already a test which tests calling `PhpOffice\PhpWord\Shared\ZipArchive::open()` with no second arg passed (i.e. `null` passed). The version of phpunit in the project means I can't run the test suite locally to verify that the 8.1 deprecation has been fixed.

I'm also not able to run the composer script because `composer install` installed an old version of phpunit that won't run on PHP 8.

### DIFF
--- a/src/PhpWord/Shared/ZipArchive.php
+++ b/src/PhpWord/Shared/ZipArchive.php
@@ -133,6 +133,13 @@ class ZipArchive
 
         if (!$this->usePclzip) {
             $zip = new \ZipArchive();
+
+            // PHP 8.1 compat - passing null as second arg to \ZipArchive::open() is deprecated
+            // passing 0 achieves the same behaviour
+            if ($flags === null) {
+                $flags = 0;
+            }
+
             $result = $zip->open($this->filename, $flags);
 
             // Scrutizer will report the property numFiles does not exist


PowerKiKi:
It's been released as https://github.com/PHPOffice/PHPWord/releases/tag/0.18.3

PR #2227, opened 2022-05-18
Template processor hyperlinks

velPL:
### Description

Fixes # (issue)

### Checklist:


VeryColdAir:
I tried to implement this and I got it to work, but it's using the default font/style for the link instead of the existing font. Is there a way to get it to keep the font the same as the text the link is replacing?

velPL:
@VeryColdAir you should be able to set all the styling you want as argument 2 is a Link class element, which can have settings as described here https://phpword.readthedocs.io/en/latest/elements.html#links

PR #2233, opened 2022-05-26
Added possibility od defining own Heading detection regular expression in Word2007 document reader

velPL:
### Description

This pull request adds a feature which allows to set own regular expression for Headings detection in ``\PhpOffice\PhpWord\Reader\Word2007\AbstractPart::getHeadingDepth()``. This is useful for Word documents produced in non-English versions of the software where default styling naming convention differs, resulting in all heading-styled elements becoming regular paragraphs in PHPWord. One can set this in php-word.ini file globally or use ``PhpOffice\PhpWord::setHeadingDepthRegexp()`` anywhere in application

### Checklist:

- [x] I have run `composer run-script check --timeout=0` and no errors were reported (there is one complaint on \PhpWord\Element\TOC.php which has already been there)

### DIFF
--- a/docs/general.rst
+++ b/docs/general.rst
@@ -151,6 +151,24 @@ default font by using the following two functions:
     $phpWord->setDefaultFontName('Times New Roman');
     $phpWord->setDefaultFontSize(12);
 
+Heading detection regexp
+~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


velPL:
Updated the feature to allow setting at which position of the regex matches actual heading number is expected to be set for the detection regex provided by user

PR #2280, opened 2022-08-19
gracefully handle absent href attribute when parsing links

### LINKED ISSUE #1562: A tag without href causes error in Rels
The following causes an error "Invalid parameters passed" in Rels

`HTMLParser::addHtml($section,'<a title="Link to Foo">Foo</a>',false);`

The link above is nonsense logically (no href and no name makes it pointless in HTML) but it's still valid HTML. The lack of href causes an exception rather than (for example) rendering it as unlinked text.

asmundstavdahl:
### Description

Previously, an 'Invalid parameters passed' error would be emitted when parsing an "a" element without an "href" attribute. This fix lets href-less links through. Thanks to @santerref of #1562 for this fix.

Fixes #1562

### Checklist:

    - There were errors, but they are not caused by the changes in this PR and should be fixed separately IMHO
    - I could not find a suitable place to document it, and also this is a fix for existing functionality and should not need to be documented.

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -1016,7 +1016,11 @@ class Html
         }
         $styles['font'] = self::parseInlineStyle($node, $styles['font']);
 
-        if (strpos($target, '#') === 0) {
+        if (empty($target)) {
+            $target = '#';
+        }
+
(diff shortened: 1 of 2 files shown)


Progi1984:
Superseeded by #2663

PR #2308, opened 2022-10-07
Call static instead of self on protected method

fbclol:
### Description

call static instead of self on protected method

Fixes # (issue)

### Checklist:


Progi1984 (approved):

Progi1984:
Thanks @fbclol

PR #2318, opened 2022-10-25
Use highest quality for JPEG and PNG images

### LINKED ISSUE #2317: Use highest quality for JPEG and PNG images
### Is your feature request related to a problem? Please describe.

When inserting images in a Word document I sometimes find the _quality_ of these images poor. Especially for JPEG files. I discovered that in src/PhpWord/Element/Image.php the function `imagejpeg()` is used for creating an image in the document. That function has a parameter called `$quality`. According to https://www.php.net/imagejpeg, a quality of about 75 is used as default for generating the jpeg. I find this value to be too low.

### Describe the solution you'd like
The functions `imagejpeg()` and `imagepng()` have a parameter called `$quality`. I'd like PHPWord to pass a value for that parameter and default it to the maximum quality.
- For `imagejpeg()`, the value for the maximum quality is `100`.
- For `imagepng()`, the value for the maximum quality is `0` (see https://www.php.net/imagepng).

### Describe alternatives you've considered

None.

### Additional context
Before (JPEG quality of 75 in Word document):
jpeg-image-quality-75

After (JPEG quality of 100 in Word document):
jpeg-image-quality-100

MegaChriz:
### Description

This sets the quality of JPEG and PNG files to their maximum. It does this by:
- defining a property called `$imageQuality` and a method called `getImageQuality()` to PhpOffice\PhpWord\Element\Image;
- setting the property `$imageQuality` to `100` in case of image/jpeg or image/jpg and set it to `0` in case of image/png;
- PhpOffice\PhpWord\Media and PhpOffice\PhpWord\Writer\AbstractWriter making use of the 'imageQuality' property.

I also added test coverage for PhpOffice\PhpWord\Element\Image::getImageQuality().

Fixes #2317

### Checklist:

### DIFF
--- a/src/PhpWord/Element/Image.php
+++ b/src/PhpWord/Element/Image.php
@@ -89,7 +89,7 @@ class Image extends AbstractElement
     /**
      * Image function.
      *
-     * @var string
+     * @var null|callable(resource): void
      */
     private $imageFunc;
 
@@ -99,6 +99,16 @@ class Image extends AbstractElement
      * @var string
      */
     private $imageExtension;
+
+    /**
+     * Image quality.
+     *
+     * Functions imagepng() and imagejpeg() have an optional parameter for
+     * quality.
+     *
+     * @var null|int
+     */
+    private $imageQuality;
 
     /**
      * Is memory image.
@@ -249,11 +259,19 @@ class Image extends AbstractElement
     /**
      * Get image function.
      *
-     * @return string
+     * @return null|callable(resource): void
      */
-    public function getImageFunction()
+    public function getImageFunction(): ?callable
     {
         return $this->imageFunc;
+    }
+
+    /**
+     * Get image quality.
+     */
+    public function getImageQuality(): ?int
+    {
+        return $this->imageQuality;
     }
 
     /**
@@ -317,20 +335,13 @@ class Image extends AbstractElement
     }
 
     /**
-     * Get image string data.
-     *
-     * @param bool $base64
-     *
-     * @return null|string
-     *
-     * @since 0.11.0
+     * Get image string.
      */
-    public function getImageStringData($base64 = false)
+    public function getImageString(): ?string
     {
         $source = $this->source;
         $actualSource = null;
         $imageBinary = null;
-        $imageData = null;
         $isTemp = false;
 
         // Get actual source from archive image or other source
@@ -367,7 +378,8 @@ class Image extends AbstractElement
                 imagesavealpha($imageResource, true);
             }
             ob_start();
-            call_user_func($this->imageFunc, $imageResource);
+            $callback = $this->imageFunc;
+            $callback($imageResource);
             $imageBinary = ob_get_contents();
             ob_end_clean();
         } elseif ($this->sourceType == self::SOURCE_STRING) {
@@ -379,20 +391,36 @@ class Image extends AbstractElement
                 fclose($fileHandle);
             }
         }
-        if ($imageBinary !== null) {
-            if ($base64) {
-                $imageData = chunk_split(base64_encode($imageBinary));
-            } else {
-                $imageData = chunk_split(bin2hex($imageBinary));
-            }
-        }
 
         // Delete temporary file if necessary
         if ($isTemp === true) {
             @unlink($actualSource);
         }
 
-        return $imageData;
+        return $imageBinary;
+    }
+
+    /**
+     * Get image string data.
+     *
+     * @param bool $base64
+     *
+     * @return null|string
+     *
+     * @since 0.11.0
+     */
+    public function getImageStringData($base64 = false)
+    {
+        $imageBinary = $this->getImageString();
+        if ($imageBinary === null) {
+            return null;
+        }
+
+        if ($base64) {
+            return chunk_split(base64_encode($imageBinary));
+        }
+
+        return chunk_split(bin2hex($imageBinary));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/PhpWord/Element/Image.php
```diff
@@ -367,7 +387,11 @@ public function getImageStringData($base64 = false)
                 imagesavealpha($imageResource, true);
             }
             ob_start();
-            call_user_func($this->imageFunc, $imageResource);
+            if (!empty($this->imageQuality)) {
```
PowerKiKi, line 390:
```sh
php -r 'var_dump(empty(0));'

bool(true)
```

So this condition will not apply an imageQuality of 0, instead it will use the second call without specifying a quality.

Also why test for the present of imageQuality at all ? why not pass it in all cases ?

PowerKiKi (changes requested):

src/PhpWord/Element/Image.php
MegaChriz: Only `imagejpeg()` and `imagepng()` accept a parameter for quality (and for both that happens to be the third parameter). `imagegif()` doesn't have that so you would else pass more parameters to that function than that it defines.
While PHP does allow additional parameters to be passed to a function, I think that we should avoid doing that if it is not necessary.

MegaChriz: Also, I'm not sure if the default value of a PHP function is kept if you explicitly pass `NULL` to it.

MegaChriz: The default value for parameter `$quality` in `imagejpeg()` is `-1`: https://www.php.net/imagejpeg

PowerKiKi: damned, I was not aware of how `$this->imageFunc` worked :roll_eyes: 

What I would do is inside `setFunctions()`, instead of affecting a string, create and affect an anonymous function. So we would have anonymous functions for each type of image but they can all be called in this same way.

More or lesse something like:

```php
$this->imageFunc = function ($resource) { imagepng($resource, null, $this->imageQuality); };
// OR 
$this->imageFunc = function ($resource) { imagejpeg($resource, null, $this->imageQuality); };
```

Idem for imageCreateFunc. I realize this is out of scope of this PR, but if you are willing to do that, great ! Otherwise let me know.

src/PhpWord/Element/Image.php
```diff
@@ -256,6 +266,16 @@ public function getImageFunction()
         return $this->imageFunc;
     }
 
+    /**
+     * Get image quality.
+     *
+     * @return null|int
+     */
+    public function getImageQuality()
```
Progi1984, line 274:
```suggestion
    public function getImageQuality(): ?int
```

Progi1984 (approved):
A ms

src/PhpWord/Element/Image.php
MegaChriz: @PowerKiKi
I would love to refactor it in the way that you are suggesting, but I see that I'm not able to get to that right away.

But I see that the current implementation has its limits and accidentally works just because parameter `$quality` is at both positions for `imagejpeg()` and `imagepng()`.

PowerKiKi:
@MegaChriz, I pushed the changes I had in mind. Would you mind testing that it still behave as you expect, please ?

MegaChriz:
@PowerKiKi Great! I plan to test it this Wednesday.

src/PhpWord/Element/Image.php
```diff
@@ -503,31 +531,45 @@ private function setFunctions(): void
         switch ($this->imageType) {
             case 'image/png':
                 $this->imageCreateFunc = $this->sourceType == self::SOURCE_STRING ? 'imagecreatefromstring' : 'imagecreatefrompng';
-                $this->imageFunc = 'imagepng';
+                $this->imageFunc = function ($resource): void {
+                    imagepng($resource, null, $this->imageQuality);
+                };
                 $this->imageExtension = 'png';
+                $this->imageQuality = 0;
```
lucasnetau, line 538: PNG $quality is the zlib compression level, it is a lossless value. $filters is what enables any lossy compression but it is disabled by default. Setting PNG quality to 0 disables all compression resulting in larger files with no quality benefit.

The default for https://www.php.net/imagepng is -1 which in-turn defaults to the zlib compression default which from the zlib documentation is a quality value of 6 

> The compression level must be Z_DEFAULT_COMPRESSION, or between 0 and 9: 1 gives best speed, 9 gives best compression, 0 gives no compression at all (the input data is simply copied a block at a time). Z_DEFAULT_COMPRESSION requests a default compromise between speed and compression (currently equivalent to level 6).

Suggest changing the PNG quality value to -1 not 0 in this patch.

MegaChriz:
With the latest changes images fail to render in my case:
no-image

I'm using the following code for parsing an image (because the image is not publicly available, a cookie is needed to retrieve it):
```php
protected static function parseImage($node, $element)
{
    $src = $node->attributes->getNamedItem('src')->nodeValue;

    $image_attr = [];
    $attributes = [
      'width',
      'height',
    ];
    foreach ($attributes as $key) {
      $attribute = $node->attributes->getNamedItem($key);
      if ($attribute) {
        $image_attr[$key] = $attribute->nodeValue;
      }
    }
    $context = stream_context_create([
      'http' => [
        'method' => 'GET',
        'header' => [
          'Cookie: ' . $_SERVER['HTTP_COOKIE']."\r\n",
        ],
      ],
    ]);

    try {
      return $element->addImage(file_get_contents($src, FALSE, $context), $image_attr);
    }
    catch (InvalidImageException $e) {
      // Log invalid image exceptions.
      watchdog_exception('panelsreport', $e);
    }
    return NULL;
}
```

When trying to debug this, I see that previously the image generation happened in `AbstractWriter::addFilesToPackage()`. Now I assume it should happen in `Image::getImageString()` but in there `$this->sourceType` happens to be equal to `'string'`, not `'gd'`. There was no such check in `AbstractWriter::addFilesToPackage()`, apparently that used the image generation functions regardless of the source type.

PowerKiKi:
Released as [1.1.0](https://github.com/PHPOffice/PHPWord/releases/tag/1.1.0)

Issue #2309, opened 2022-10-07
Problem with TemplateProcessor.php


Ervdu13:
The functions "replaceBlock" and "deleteBlock" don't work.

in the function "replaceBlock" the preg_match isn't correct.
if I replace  
'/(<\?xml.*)(<w:p.*>\${' . $blockname . '}<\/w:.*?p>)(.*)(<w:p.*\${\/' . $blockname . '}<\/w:.*?p>)/is' 
by
 '/(.*((?s)<w:p\b(?:(?!<w:p\b).)*?\${' . $blockname . '}<\/w:.*?p>))(.*)((?s)<w:p\b(?:(?!<w:p\b).)[^$]*?\${\/' . $blockname . '}<\/w:.*?p>)/is'
 (like in the function "cloneBlock"), the 2 functions are Ok

FournyP:
Hello,

I think I have the same problem, when I use these functions in my code, the returned word file is corrupted.

I have try your solution @Ervdu13, the file is no more corrupted but the replace didn't replace, same comportement as a delete

FournyP:
I have found a solution to avoid the problem,

cloneBlock work correctly, so if you want to delete the block, just clone it to 0 like this :
```php
$templateProcessor->cloneBlock('myBlockName', 0, true);
```

And for replace, you have to put a variable into your template like this :
```
${myBlockName}
${myVar}
${/myBlockName}
```
And do this in your code :
```php
$templateProcessor->cloneBlock('myBlockName', 1, true, false);
$templateProcessor->setValue('myVar', <Your value here>);
```

PR #2221, opened 2022-05-09
Fix #1970

tontonsb:
### Description

Fixes `deleteBlock` by using `cloneBlock` instead of `replaceBlock`, because it has a better regex.

Fixes #1970

### Checklist:

- [ ] I have run `composer run-script check --timeout=0` and no errors were reported — I made the change via GitHub UI and hope that the GitHub workflow will run the tests once I make this PR
- [ ] The new code is covered by unit tests (check build/coverage for coverage report) — no new lines were added
- [ ] I have updated the documentation to describe the changes — no features were added or changed, this is a bug fix

Issue #2051, opened 2021-04-09
DeleteBlock / ReplaceBlock wrong regex


philip-sorokin:
The following regex does not match:

`/(<\?xml.*)(<w:p.*>\${' . $blockname . '}<\/w:.*?p>)(.*)(<w:p.*\${\/' . $blockname . '}<\/w:.*?p>)/is`

UPD. Word 2007 with sections puts the content in child XML elements, so, it's not possible to remove it with a regular expression.

philip-sorokin:
To avoid backtracking you can replace the content between parent rows in 2 steps:

`		$this->tempDocumentMainPart = preg_replace('#(\${/?' . $blockname . '})(.*?</w:p>)#is', '$2$1', $this->tempDocumentMainPart);`

`		$this->tempDocumentMainPart = preg_replace('#\${' . $blockname . '}.*\${/' . $blockname . '}#is', '', $this->tempDocumentMainPart);`

The first regex puts the opening and closing block markers **after** the parent's XML tag. The second regex performs replacing the content between tags.

philip-sorokin:
Is there a method to delete block markers that are not involved in replacing?

PR #2319, opened 2022-10-26
refactor: add wrapper around `preg_match`

drupol:
Hi,

### Description

The default behavior of [`preg_match`](https://www.php.net/manual/en/function.preg-match.php) is to silent any parsing error.
Therefore, I created a this wrapper is here to make sure that it properly throws an exception in case of any `preg_match` issue.

The origin of this PR comes from an issue we had at work. We were trying to delete a block in a .docx file (75Kb) and nothing was happening, despite the fact that the method issued correctly.
We made an introspection of the `document.xml` and discovered that the method [`TemplateProcessor::deleteBlock`](https://github.com/PHPOffice/PHPWord/blob/develop/src/PhpWord/TemplateProcessor.php#L855) doesn't work, it seems that the regex is not valid (_there are already a couple of issues about that: #2051, #2221, #2309, #2157_), we are now using [`TemplateProcessor::cloneblock`](https://github.com/PHPOffice/PHPWord/blob/develop/src/PhpWord/TemplateProcessor.php#L791) instead.
Then, we narrowed down the issue to the `preg_match` function and succeeded to reproduce the issue with a simple xml document.
This is were I discovered the function [`preg_last_error`](https://www.php.net/manual/en/function.preg-last-error.php) and noticed that the issue was related to the `pcre.backtrack_limit` PHP directive. Once we updated that directive from `1000000` to `2000000` it worked.

We lost less or more half a day on this issue, and this PR should help some people to see clearer when something goes wrong.

Could you please let me know if this PR is legit and if I should invest more time in it?

Nice to have: ideally, it would be a good think to invest time into an external contrib library that does `preg_match` correctly with the proper Exception handling.

Thanks

### Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,6 +61,7 @@
         "php": "^7.1|^8.0",
         "ext-dom": "*",
         "ext-json": "*",
+        "ext-pcre": "*",
         "ext-xml": "*",
         "laminas/laminas-escaper": ">=2.6"
     },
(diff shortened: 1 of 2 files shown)


drupol:
Closing this PR since no feedback after 3 months.

kovalovme:
BUMP

Issue #2330, opened 2022-11-21
vMerge value is Empty in MS Word 2021


kernusr:
### Describe the Bug

In MS Word 2021 docx files save with empty `w:val` of `w:vMerge` when expected value "continue".

### Steps to Reproduce

1. Create new document in MS Word 2021
2. Create table in document
3. Merge multiple cells vertically
4. Save the document and check word\document.xml

I tried saving:
- as Word Document (*.docx)
- as Strickt Open XML Document (*.docx)

In both cases I get the described problem

When I save in Libre Office the problem doesn't reproduce

What does this have to do with phpword?
When I want to save a file in html or pdf I get incorrect markup

### Context

Please fill in your environment information:

- PHPWord Version: 1.0

kernusr:
I can suggest a solution.

In method `readStyleDefs` for argument `styleDefs` add fifth parameter "default value"
https://github.com/PHPOffice/PHPWord/blob/0af4ca1f6e52be05ac3f87fcb68d21e85a158a98/src/PhpWord/Reader/Word2007/AbstractPart.php#L624-L653

Modify the code as follows
```php
    protected function readStyleDefs(XMLReader $xmlReader, ?DOMElement $parentNode = null, $styleDefs = [])
    {
        $styles = [];

        foreach ($styleDefs as $styleProp => $styleVal) {
            [$method, $element, $attribute, $expected, $default] = array_pad($styleVal, 4, null);

            $element = $this->findPossibleElement($xmlReader, $parentNode, $element);
            if ($element === null) {
                continue;
            }

            if ($xmlReader->elementExists($element, $parentNode)) {
                $node = $xmlReader->getElement($element, $parentNode);

                $attribute = $this->findPossibleAttribute($xmlReader, $node, $attribute);

                // Use w:val as default if no attribute assigned
                $attribute = ($attribute === null) ? 'w:val' : $attribute;
                $attributeValue = $xmlReader->getAttribute($attribute, $node) ?? $default;

                $styleValue = $this->readStyleDef($method, $attributeValue, $expected);
                if ($styleValue !== null) {
                    $styles[$styleProp] = $styleValue;
                }
            }
        }

        return $styles;
    }
```

This will allow default values to be passed for empty attributes. For example in a cell styles
https://github.com/PHPOffice/PHPWord/blob/0af4ca1f6e52be05ac3f87fcb68d21e85a158a98/src/PhpWord/Reader/Word2007/AbstractPart.php#L558

[...]

PR #2336, opened 2022-11-22
Read empty vmerge

kernusr:
### Description
In MS Word 2021 docx files save with empty w:val of w:vMerge when expected value "continue".

I made suggest a solution:
In method readStyleDefs for argument styleDefs add fifth parameter "default value". This will allow default values to be passed for empty attributes. For example in a cells

The problem is described here: https://github.com/PHPOffice/PHPWord/issues/2330

### Checklist:


Progi1984 (approved):

Progi1984:
Thanks @kernusr for your contribution.

PR #2362, opened 2022-12-30
Add support for style objects to the element constructor

hazington:
### Description

The current version does not support passing style objects to the constructor of an element. Only arrays are supported. But for a full OOP approach the element constructor should also or only support style objects. As the setNewStyle method is only used internally, the returnObject param can be ignored when passing a style element instance as styleValue via constructor.

Fixes # (issue)

Elements do not support style objects via constructor.

### Checklist:

### DIFF
--- a/docs/elements.rst
+++ b/docs/elements.rst
@@ -114,8 +114,8 @@ You can add Hyperlinks to the document by using the function addLink:
 
 - ``$linkSrc``. The URL of the link.
 - ``$linkName``. Placeholder of the URL that appears in the document.
-- ``$fontStyle``. Array or Font style object. See :ref:`font-style`.
-- ``$paragraphStyle``. Array or Paragraph style object. See :ref:`paragraph-style`.
+- ``$fontStyle``. Classname, Array or Font style object. See :ref:`font-style`.
+- ``$paragraphStyle``. Classname, Array or Paragraph style object. See :ref:`paragraph-style`.
 
 Preserve texts
 ~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


hazington:
I'll cover this one with unit tests until the end of this week.

hazington:
@Progi1984 

UnitTests added, documentation modified. You're welcome!

hazington:
While I was working on this issue I realized that the style behavior should be revised. The constructor and set style methods should only support the actual style object. The array support should be abandoned. Instead a static fromArray method should be added to each style.

Before:
```php 
$cell = new cell(100, [
   // some style
]);
```

After: 
```php
$cell = new cell(100, CellStyle::fromArray([
   // some style
]);
```

Since this would be a breaking change, I decided to go for the workaround committed here to at least add support for style objects for elements like cell. I've also noticed that the style behavior is quite different across the various element types. The text object for example always uses the same object passed via constructor. Other elements prefer cloning.

I think that we should also add a merge method to element styles and a dedicated className array. All elements should always return a style object via getStyle() method.

```php
[...]
public function addClassName(string $className): void;
public function hasClassName(string $className): bool;
public function clearClassNames(): void;
public function hasClassNames(): bool;
public function getClassNames(): array;
public function getStyle(): <StyleObject>;
[...]
```

With the merge method something like this could happen in the renderer:

```php
$style = $element->getStyle();

foreach ($phpWord->getStyles($element->getStyleNames() as $classStyle) {
    if ($classStyle::class === $style::class) {
       $style->merge($classStyle);
   }
}
```

This would allow adding multiple class names to an object but also overwrite specific attributes it. My private html2word library does support multiple classes in a similar way, but I don't know how Word files are actually rendered. If Word only supports one class name per element, then passing multiple classes to an element doesn't make sense. But even then, the set style method should only support `string|<style-object>'.

PR #2363, opened 2022-12-30
Add support for fractional font sizes

hazington:
### Description

Word supports fractional font sizes like 8.5pt, but PHPWord doesn't.

Fixes # (issue)

That fractional font size ist not supported as default font size, while font style elements support fractional font size.

### Checklist:

### DIFF
--- a/src/PhpWord/Settings.php
+++ b/src/PhpWord/Settings.php
@@ -2,10 +2,8 @@
 /**
  * This file is part of PHPWord - A pure PHP library for reading and writing
  * word processing documents.
- *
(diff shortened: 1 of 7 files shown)


hazington:
@Progi1984 

I've added type hinting to the settings class and improved the set default font size method. It now supports both, int and float font sizes. I've also updated the unit tests to cover the new supported fractional font sizes. For an unknown reason the rebase command also added the new textbox code of my other patch that has already been merged to the master branch.

Progi1984 (approved):

PR #2364, opened 2022-12-30
Add background color support for text box element for Word writer

hazington:
### Description

It is now possible to fill the background color of text box elements (only Word2007 writer!). There is no documentation for the Textbox element, so I can't add the changes.

Fixes # (issue)

No bugfix.

### Checklist:

### DIFF
--- a/src/PhpWord/Style/TextBox.php
+++ b/src/PhpWord/Style/TextBox.php
@@ -65,6 +65,33 @@ class TextBox extends Image
      * @var string
      */
     private $borderColor;
+    
+    /**
+     * background color
+     *
+     * @var string
+     */
+    private $bgColor;
+
+    /**
+     * Set background color
+     *
+     * @param string $value
+     */
+    public function setBgColor($value = null)
+    {
+        $this->bgColor = $value;
+    }
+
+    /**
+     * Get background color
+     *
+     * @return string
+     */
+    public function getBgColor()
+    {
+        return $this->bgColor;
+    }
 
     /**
      * Set margin top.
--- a/src/PhpWord/Style/TextBox.php
+++ b/src/PhpWord/Style/TextBox.php
@@ -65,7 +65,7 @@ class TextBox extends Image
      * @var string
      */
     private $borderColor;
-    
+
     /**
      * background color.
      *
@@ -78,7 +78,7 @@ class TextBox extends Image
      *
      * @param string $value
      */
-    public function setBgColor($value = null)
+    public function setBgColor($value = null): void
     {
         $this->bgColor = $value;
     }
--- a/src/PhpWord/Style/TextBox.php
+++ b/src/PhpWord/Style/TextBox.php
@@ -9,6 +9,7 @@
  * contributors, visit https://github.com/PHPOffice/PHPWord/contributors.
  *
  * @see         https://github.com/PHPOffice/PHPWord
+ *
  * @license     http://www.gnu.org/licenses/lgpl.txt LGPL version 3
  */
 
@@ -72,19 +73,14 @@ class TextBox extends Image
 
     /**
      * Set background color.
-     *
-     * @param null|string $value
-     * @return void
      */
-    public function setBgColor(string $value = null): void
+    public function setBgColor(?string $value = null): void
     {
         $this->bgColor = $value;
     }
 
     /**
      * Get background color.
-     *
-     * @return null|string
      */
     public function getBgColor(): ?string
     {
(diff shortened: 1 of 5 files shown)


src/PhpWord/Style/TextBox.php
```diff
@@ -66,6 +66,33 @@ class TextBox extends Image
      */
     private $borderColor;
 
+    /**
+     * background color.
+     *
+     * @var string
```
Progi1984, line 72:
```suggestion
     * @var string|null
```

src/PhpWord/Style/TextBox.php
```diff
@@ -66,6 +66,33 @@ class TextBox extends Image
...
+     */
+    private $bgColor;
+
+    /**
+     * Set background color.
+     *
+     * @param string $value
+     */
+    public function setBgColor($value = null): void
```
Progi1984, line 81:
```suggestion
    public function setBgColor(string $value = null): self
```

src/PhpWord/Style/TextBox.php
```diff
@@ -66,6 +66,33 @@ class TextBox extends Image
...
+    {
+        $this->bgColor = $value;
```
Progi1984, line 83:
```suggestion
        $this->bgColor = $value;
        
        return $this;
```

src/PhpWord/Style/TextBox.php
```diff
@@ -66,6 +66,33 @@ class TextBox extends Image
...
+     * background color.
+     *
+     * @var string
+     */
+    private $bgColor;
+
+    /**
+     * Set background color.
+     *
+     * @param string $value
```
Progi1984, line 79:
```suggestion
     * @param string $value
     * @return self
```

src/PhpWord/Style/TextBox.php
```diff
@@ -66,6 +66,33 @@ class TextBox extends Image
...
+    {
+        $this->bgColor = $value;
+    }
+
+    /**
+     * Get background color.
+     *
+     * @return string
+     */
+    public function getBgColor()
```
Progi1984, line 91:
```suggestion
    public function getBgColor(): ?string
```

Progi1984 (commented):
@hazington Some feedbacks to apply before merge

Progi1984:
@hazington Do you think you can add unit tests ?

hazington:
@Progi1984 

Yes, I can cover this code with an unit test. By the way, this code is based on version 0.18.x and copied from another style element of PHPWord that already supported setting the background color. So this is not my coding style, it simply followed the old project's style.

I appreciate your feedback as I personally prefer coding in PHP 8.1, with strict_types=1 and based on PSR-12. By any chance, is refactoring based on PSR-12 and at least strong typed coding on the road map? That would make supporting PHPWord so much easier! :)

src/PhpWord/Style/TextBox.php
hazington: Return $this in a setter method is not used in any other styling class' setter method. Doing this for only one class or even worser for only one class method is very inconsistent. I think this should be changed for all setters in version 1.1.x.

hazington:
So, the requested unit tests have been added and the code style fixed according to the php-cs-fixer configuration by running the fix script from composer.json.

hazington:
I've some trouble with the PHP 7.1 and PHP 7.2 coverage. Using "assertRegExp" will fail the CI / coverage and using the "assertMatchesRegularExpression" will let the coverage for PHP 7.1 and 7.2 fail. Any solution here? Since PHP 7.4 is officially EOL, dropping coverage for at least PHP 7.1, 7.2 and 7.3 should be considered.

https://github.com/PHPOffice/PHPWord/pull/2364/commits/679a738c426ff498bd1b4319c23ad1ef438d5e10

hazington:
This workaround seems to solve the CI coverage issue:

[53d34fd](https://github.com/PHPOffice/PHPWord/pull/2364/commits/53d34fd049026752a13183891250bd3d91c482c0)

```php
        // Try to address CI coverage issue for PHP 7.1 and 7.2 when using regex match assertions
        if (method_exists(static::class, 'assertRegExp')) {
            self::assertRegExp('/z\-index:\-[0-9]*/', $style);
        } else {
            self::assertMatchesRegularExpression('/z\-index:\-[0-9]*/', $style);
        }
```

Progi1984 (approved):

Progi1984:
Thanks @hazington

PR #2398, opened 2023-05-04
Update Language.php  FR_CH, DE_CH & IT_CH

SarahTrees:
Microsoft knows the variations for DE, FR and IT in Switzerland.
3 language codes added:

Add FR_CH
Add DE_CH
Add IT_CH 

https://learn.microsoft.com/en-us/previous-versions/tn-archive/cc287874(v=office.12)


SarahTrees:
does anyone know why the checks are not successful?
Since 2023, this seems to be a problem with all pull requests. Who has an idea?

PowerKiKi:
Thanks !

PR #2404, opened 2023-05-30
Lock all dependencies

PowerKiKi:
### Description

We must have `composer.lock` as part of our repository. Otherwise, the tools used for development change and break things unexpectedly.

Currently, this is the case with PHPUnit 10 that introduces quite a few breaking changes. We cannot `composer install` and hope that we will get a set of compatible dependencies. Instead, we must lock our tools and upgrade them only when we want/need to upgrade them specifically, in a dedicated commit, and not part of a totally unrelated work, because it just so happenned that PHPUnit released something.

`composer.lock` is based on the oldest PHP version that is officially supported by PHP core. Currently, that is 8.0. This typically gives us compatibility out of the box for 8.0, 8.1 and 8.2 with a single lock file.

EOL PHP versions, that we unfortunately keep supporting in PHPWord, are "the special cases", and as such they will generate a new lock files in CI, because we cannot have a single lock files that spread over such as long time (almost 4 years for PHP 7.1) with major breaking change in the language in between.

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.1"
+          - "7.1.3"
           - "7.2"
           - "7.3"
           - "7.4"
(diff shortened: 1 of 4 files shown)


PowerKiKi:
This typically solves issues as seen in #2401 were a trivial PR suddenly break CI without reasonable reasons. So by doing this we lower the barrier for contributions.

Issue #2410, opened 2023-06-09
CloneBlock, replaceBlock, and deleteBlock not working in template docx


MuhammadSyahmi98:
### Describe the Bug

I am experiencing issues with the cloneBlock, replaceBlock, and deleteBlock methods in the PHPWord library. These methods are used to manipulate template DOCX files by cloning, replacing, or deleting specific blocks of content. However, they are not functioning as expected.

### Steps to Reproduce

Prepare a template DOCX file with the following content:

```${block_name}
Customer: ${customer_name}
Address: ${customer_address}
${/block_name}
```

In the controller or PHP script, use the following code:
```
$templateProcessor->cloneBlock('block_name', 3, true, true);
```

### Expected Behavior

The cloneBlock method should clone the block_name block three times in the template DOCX file, resulting in multiple repetitions of the block.

### Current Behavior

The cloneBlock method does not clone the block as expected. No changes are made to the template DOCX file. Still same like original template

### Additional Information

- The issue persists even when using the simple code provided in the PHPWord documentation.
- The setValue and setValues methods work correctly, allowing the replacement of individual placeholders in the template.
- This problem occurs specifically with the cloneBlock, replaceBlock, and deleteBlock methods.
- The issue is present when using PHP 8.1.
- I have verified that I am using the latest version of the PHPWord library.
- No errors or exceptions are thrown during the execution of the code.

### Context

Please fill in your environment information:

- PHP Version: 8.1
- PHPWord Version: 1.1.0
- Laravel Version: 10

MuhammadSyahmi98:
#341 solution from @liborm85 is working. For my case need to change the preg_match code. If have problem will update soon. Hope someone can fix this issue without need to override existing code

Create new controller as TemplateProcessorMod extends TemplateProcessor

```
public function cloneBlock($blockname, $clones = 1, $replace = true, $indexVariables = false, $variableReplacements = null)
    {
        $xmlBlock = null;
        $matches = array();
        preg_match(
            '/(\${' . $blockname . '})(.*?)(\${\/' . $blockname . '})/is',
            $this->tempDocumentMainPart,
            $matches
        );

        if (isset($matches[3])) {
            $xmlBlock = $matches[2];
            if ($indexVariables) {
                $cloned = $this->indexClonedVariables($clones, $xmlBlock);
            } elseif ($variableReplacements !== null && is_array($variableReplacements)) {
[...]

PR #2261, opened 2022-07-15
Allow to customize macro delimiters in TemplateProcessor

ismail1432:
### Description

Hello 👋 

Congratulations on this huge project 🔥 

I'm glad to open a pull request with new additions where the goal is to allow developers to customize macros delimiters with 3 methods:
- `setMacroOpeningChars`
- `setMacroClosingChars`
- `setMacroChars`

You can see #2251 for more details.

💡 I ran `composer run-script check --timeout=0` and got an output that is not related to the PR: `CamelCasePropertyName  The property $TOCStyle is not named in camelCase.`

✅  I duplicated all tests related to the TemplateProcessor with the customization of macros and it works like a charm.

I let you check the PR let me know if something is wrong or if you need more information.

### Checklist:

- [x] I have run `composer run-script check --timeout=0` and no errors were reported
- [x] The new code is covered by unit tests (check build/coverage for coverage report)
- [x] I have updated the documentation to describe the changes
- [x]  Add more tests

Issue #1836, opened 2020-03-10
TemplateProcessor::cloneBlock() doesn't work with PHP v7.3+


cedrictailly:
### Describe the Bug

It seems the regex used to search the block doesn't work with a template we use, the problem seems to be in relation with the lazy quantifiers, large strings and PHP v7.3 and superior, it works well with PHP v7.2.

### Steps to Reproduce

I can't send the template because it's confidential, but :
 - When I test lazy quantifiers on simple tests nothing is wrong (PHP v7.3 + v7.4).
 - When I test with exact same data minus unuseful tags it works too (PHP v7.3 + v7.4).
 - When I test with the full file content there is no match with PHP v7.3+ but a match with v7.2.

### Expected Behavior

The blocks have to be cloned...

### Current Behavior

...but tags remain visibles.

### Context

- PHP Version: 7.3+
- PHPWord Version: 0.17.0

cHeeSaW:
We expirienced similiar issues and dived a bit deeper into it. It seems not to be a PHP-Version Problem but rather a Problem with the PCRE Version behind. Running on PCRE 10.33 works, PCRE 10.34 fails.

cedrictailly:
I met other problematic cases when the matches returned by **preg_match()** is too big (I suppose but it's very probable), I'm a big newbie with the Word file format but I've rewritten this function with what I've seen in my cases, the function to replace is in file **/src/PhpWord/TemplateProcessor.php** and here is my alternative which uses **mb_\*str\*()** functions and works at least in all of my cases :

```php
public function cloneBlock($blockname, $clones = 1, $replace = true, $indexVariables = false, $variableReplacements = null)
{
    $idx_tag = mb_strpos($this->tempDocumentMainPart, '${' . $blockname . '}');

    if ( $idx_tag === false )
      return null;

    $idx_start = mb_strrpos(mb_substr($this->tempDocumentMainPart, 0, $idx_tag), '<w:p '                 );
    $idx_end   = mb_strpos(          $this->tempDocumentMainPart,               '${/' . $blockname . '}');

    if ( $idx_start === false || $idx_end === false )
      return null;

    $idx_end = mb_strpos($this->tempDocumentMainPart, '</w:p>', $idx_end);

    if ( $idx_end === false )
      return null;

    $idx_end += 6;

    $what = mb_substr($this->tempDocumentMainPart, $idx_start, $idx_end - $idx_start);

    // --- //

[...]

PR #2448, opened 2023-08-29
fix block handling in template processor

### LINKED ISSUE #2444: Allow to Separate Variables and Blocks
### Is your feature request related to a problem? Please describe.

In TemplateProcessor.php, the following characters are defined

    protected static $macroOpeningChars = '${';

    protected static $macroClosingChars = '}';

And are used for:
- Cloning Blocks
- Replacing Variables (through ensureMacroCompleted)

Then the cloning blocks macro should be replaced by '' if only used to manage blocks.

It is even more annoying if you want to use a variable that is used inside a block to duplicate a block...

### Describe the solution you'd like

Allow to define different characters for variables and blocks

### Describe alternatives you've considered

No generic way to distinguish between blocks and variable in an automated way, with no assumption in the document

### Additional context

My goal is to define the way the template is filled from the template itself, and not from any PHP code.

glaszig:
- respect configured macro delimiters in block content (introduced in #2261)
- more relaxed but reliable regexp to detect blocks
- support multiple blocks with same name

maybe fixes #2444, #2410
related: #1836, #1354 

### Checklist:


glaszig:
@coveralls don't be that pedantic.

thomasb88:
Why don't you use the 
findContainingXmlBlockForMacro function ?

In some way it has the same logic as your regexp, but your regexp seems complex for big files.

Whereas findContainingXmlBlockForMacro use first findMacro to find it straight, then from the position found, search for the first opening w:p.

I had reflexion about this problem, and my conclusion is not everything should be done by regexp to manage well complexity (else the regexp would use Lookahead or Lookbehind feature that are having complexity far more than O(n)).

I also customized a little bit this function, like

/**
     * Clone a block.
     *
     * @param string $blockname
     * @param int $clones How many time the block should be cloned
     * @param bool $replace
     * @return string|null
     */
    public function cloneBlock($blockname, $clones = 1, $replace = true, $separator = 'linebreak')
    {
	    // Separator
		switch($separator){
		    case "pagebreak":
			    $objectClass = 'PhpOffice\\PhpWord\\Element\\' . 'PageBreak';
			    $xml_obj = new $objectClass();
			    $xml_separator = $this->getXmlObject($xml_obj);
			    break;
			case "linebreak":
			    $objectClass = 'PhpOffice\\PhpWord\\Element\\' . 'TextBreak';
			    $xml_obj = new $objectClass();
			    $xml_separator = $this->getXmlObject($xml_obj);
			    break;
			default:
			    $xml_separator = '';
			    break;
		}
		// Get Block to be Cloned And Generate Cloned Data
		$rawClonedData = '';
		$xmlBlock = null;
		$where = $this->findContainingXmlBlockForMacro($blockname, 'w:p', 'Block', 1, 'regexp');
		if(false !== $where){
			$xmlBlock = $this->getSlice($where['start'], $where['end']);
            $cloned = array_fill(0, $clones, $xmlBlock);
			$rawClonedData = implode($xml_separator, $cloned);
			$rawClonedData = str_replace(self::$macroBlockOpeningChars.$blockname.self::$macroBlockClosingChars, '', $rawClonedData);
			$rawClonedData = str_replace(self::$macroBlockOpeningChars.self::$close_block_char.$blockname.self::$macroBlockClosingChars, '', $rawClonedData);
		}
		// Insert Cloned Data
		if ($replace and (strlen($rawClonedData) > 0)){
		    $this->replaceXmlBlock($blockname, $rawClonedData, 'w:p', 'Block', 1);
        }
        return $where;
    } 

And then customized findMacro to be able to compare different algorithm;
protected function findMacro($search, $offset = 0, $macro_type = 'Block', $macro_index = 1, $search_algorithm = 'regexp')
    {
        $search = static::ensureMacroCompleted($search, $macro_type);
		if('regexp' == $search_algorithm){
		    $pattern_macro = $this->regexp_delimiter.preg_quote($search, $this->regexp_delimiter).$this->regexp_delimiter.'i';
		    preg_match_all($pattern_macro, substr($this->tempDocumentMainPart, $offset), $matches, PREG_OFFSET_CAPTURE);
		    $pos = (($macro_index >= 1) and (count($matches[0]) >= $macro_index))?($offset + $matches[0][$macro_index-1][1]):-1;
		} elseif('recursive' == $search_algorithm) {
		    $pos = ($macro_index >= 1)?($offset-1*strlen($search)):false; 
            for ($i=1 ; $i <= $macro_index ; $i++){ 
                $pos = strpos($this->tempDocumentMainPart, $search, $pos+strlen($search));
                if($pos === false){
			        break;
                }			
            }
        } else {
            $pos = strpos($this->tempDocumentMainPart, $search, $offset);
        }
        return ($pos === false) ? -1 : $pos;
    }

Off course, all of this is dependant of the fixBrokenMacros that is also modified to be able to fix merge all the w:t including tha macro without breaking style and other data.

So, sorry for my long post, but the summary is that breaking the problem in function seems to be a good way to organise the TemplateProcessor processing. What do you think ?

fotrino:
Any update on this PR?

kovalovme:
BUMP

MrLexisDev:
Any update of this fix please?

PR #2491, opened 2023-10-24
check for null on $fontDefaultStyle

spatialfree:
after trying everything i could to try to absolve the offending file, i implemented this small check that get's around a null error that i encountered when converting a docx file (downloaded from google drive) to txt


Progi1984:
Superseeded by #2513.

PR #2531, opened 2023-12-17
Correct Font Size Calculated by MsDoc Reader

### LINKED ISSUE #2526: Deprecated: Implicit conversion from float 10.5 to int loses precision
### Describe the Bug

My PHP version is 8.3. When I am trying to load some file gives me some warnings for deprecated code.

### Steps to Reproduce

Please provide a code sample that reproduces the issue.

```php
<?php
$phpWord = IOFactory::load($file, 'MsDoc');
```

### Current Behavior

Deprecated: Implicit conversion from float 10.5 to int loses precision at PhpWord\Reader\MsDoc.php:1868
Deprecated: Creation of dynamic property PhpOffice\PhpWord\Shared\OLERead::$entry is deprecated (For all properties inside this file)

### Context

- PHP Version: 8.3
- PHPWord Version: 1.1.0

### Possible solution

In MsDoc.php at line 1868, use (int) to cast the result of the division to an integer explicitly.
```$oStylePrl->styleFont['size'] = dechex((int) ($operand / 2));```

To fix the second type deprecations, declare the properties in the class instead of dynamically creating it.

### PR DESCRIPTION
**Pull request by oleibman**
### Description

Fix #2526. Most of that issue has already been fixed. The one remaining problem was a deprecation message handling font size. The code used `dechex($operand / 2)`, and issued the deprecation message whenever `$operand` was odd because `dechex` is designed only for integer conversion. `$operand` is actually 2 times the point size, so it will be odd only when the point size is some integer plus half a point (no other fractions are allowed). At any rate, it seems that `dechex` should not be used here in the first place; font size is a numeric value, not a hex string.

There are many problems with MsDoc Reader at the moment. This PR is narrowly focused on the problem at hand. Its test is, at least, more detailed than the existing MsDoc Reader test, which does nothing more than confirm that read successfully creates a PhpWord object. The new test verifies that the font size is as expected, but does not validate any other aspect of the read.

Fix #2526

### Checklist:

### CODE REVIEW

File: tests/PhpWordTests/Reader/MsDocTest.php [hunk 1]
```diff
@@ -60,6 +60,22 @@ public function testLoad(): void
         self::assertInstanceOf('PhpOffice\\PhpWord\\PhpWord', $phpWord);
     }
 
+    public function testLoadHalfPointFont(): void
+    {
+        $filename = __DIR__ . '/../_files/documents/word.2526.doc';
```
Progi1984, line 65: @oleibman Could you use reader.doc ?

Progi1984 (commented):
@oleibman A small feedback ?

Could you create a 1.3.0.md file  for changelog?

tests/PhpWordTests/Reader/MsDocTest.php
oleibman: I can't use reader.doc because it does not contain any half-point fonts, so it doesn't demonstrate the problem. I could change it, but that might cause problems if anyone depends on it as-is. I agree that's pretty unlikely, so, if that's what you want me to do, I will go ahead and do that.

Progi1984: I understand. Could you rename it as reader.font-halfpoint.doc ?

oleibman: Done, and I have also added 1.3.0.md.

Progi1984 (approved):

Progi1984:
Thanks @oleibman. It's the first contribution for PhpWord 2.0

PR #2475, opened 2023-09-18
Add Support for Various Missing Features in HTML Writer

oleibman:
This PR supersedes #1814 and #2343, which had become badly out of sync due to recent changes to the repository.

Implement a number of features implemented in PhpWord, but not yet supported in PhpWord HTML Writer.

1. Use css @page and page declarations for sections.
2. Wrap sections in div, with page break before each (except first).
3. Add ability to specify generic fallback font for html (documentation change).
4. Add ability to specify handling of whitespace in html (documentation change). Currently, Word writer preserves space but HTML writer does not.
5. Support for Language, both for document overall and individual text elements.
6. Support for PageBreak for HTML (currently only PDF is supported).
7. Support for Table Border style, color, and size.
8. Support for empty paragraphs (Word writer permits, browsers generally suppress).
9. Default paragraph style should apply to all paragraphs, as well as class Normal.
10. Paragraph style should support line-height.
11. Paragraph style should support indentation.
12. Paragraph style should support page-break-before.
13. Paragraph style should not specify margin-top/bottom when spacing is null.
14. Fix #2467 - text-align:right in html is not handled correctly.

### Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

### Checklist:

- [x] I have run `composer run-script check --timeout=0` and no errors were reported
- [x] The new code is covered by unit tests (check build/coverage for coverage report)
- [x] I have updated the documentation to describe the changes

PR #2533, opened 2023-12-25
Html Reader Process Titles as Headings Not Paragraphs

### LINKED ISSUE #1692: Importing HTML Headings inserts text, not heading and export to html wrong elements
When importing from HTML via `\PhpOffice\PhpWord\Shared\Html::addHtml`, all heading-elements are handled by `parseHeading` as normal text with the extra styles `paragraph` as `Heading1,2,...`. This should be handled by the usual `addTitle`.

The above seems to work fine when exporting to Word2007, however when exporting to html, `addTitle` produced a `h1,h2,...`-element, while the former produces `<p class="Heading1,2,...">`, which is a problem, because the styles for the headings don't apply.

Also, when exporting to html, a title (depth 0) is written as `h0`, while the styles are applied to `Title`.

Finally, when exporting to html, all texts get `style="margin-top: 0; margin-bottom: 0;"`. This should not be there. They also don't get the `Normal`-class, meaning paragraph styles don't get applied.

This is a minor thing, but please also consider instead of using `* {font-family: Arial; font-size: 10pt;}` instead applying this to `body`. This way, the font is initialized and all other elements inherit it instead of re-setting it every time. This causes problems when for example using a `span` inside a `h1` (the font-size is wrong).

oleibman:
Fix #1692. Builds on work started some time ago by @0b10011, to whom primary credit is due.

Html Reader does not process the `head` section of the document, and, in particular, does not process its `style` section. It will, however, process inline styles, so 0b10011's model of adding the title as a text run (with styles) will work well once this change is applied. However, that model would not deal with the alternative method of assigning a Title Style, and just adding the title as text. In order to accommodate that, I have removed the declaration of heading font styles in the head section, and now generate them all inline in the body. This has the added benefit of being able to read the doc as html, then saving it as docx, preserving, at least in part, any user-defined font styles. Note that html does have pre-defined title styles, but docx does not.

@constip suggests in the original issue that margin top and bottom are being applied too frequently. I believe that was addressed by recently merged PR #2475. It is also suggested that the `*` css selector be dropped in favor of `body`. 2475 added the body selector. I agree that this renders the `*` selector unnecessary, and, as stated in the issue, it can cause problems. This PR drops that selector. It is also suggested that `loadHTML` be used instead of `loadXML`. This is not as easy a change as it seems, because loadHTML uses ISO-8859-1 charset rather than UTF-8, so I will not attempt that change.

### Description

Fixes # (issue)

### Checklist:


src/PhpWord/Writer/HTML/Element/Title.php
```diff
@@ -46,8 +46,14 @@ public function write()
             $writer = new Container($this->parentWriter, $text);
             $text = $writer->write();
         }
+        $css = '';
+        $style = \PhpOffice\PhpWord\Style::getStyle('Heading_' . $this->element->getDepth());
```
Progi1984, line 50: Could you define class in use function in the header of the file ?

Progi1984 (changes requested):
@oleibman Could you move changes to 2.0.0.md file, please ?

It seems that this PR is not finished. Isn't it ?

src/PhpWord/Writer/HTML/Element/Title.php
oleibman: Done, and eliminated 2 other fully qualified names as well.

src/PhpWord/Writer/HTML/Element/Title.php
```diff
@@ -18,7 +18,11 @@
 
 namespace PhpOffice\PhpWord\Writer\HTML\Element;
 
+use PhpOffice\PhpWord\Element\Title as PhpWordTitle;
```
Progi1984, line 21:
```suggestion
use PhpOffice\PhpWord\Element\Title as ElementTitle;
```

src/PhpWord/Writer/HTML/Part/Head.php
```diff
@@ -92,17 +92,16 @@ private function writeStyles(): string
...
         $defaultWhiteSpace = $this->getParentWriter()->getDefaultWhiteSpace();
         if ($defaultWhiteSpace) {
             $astarray['white-space'] = $defaultWhiteSpace;
         }
+        $bodyarray = $astarray;
 
         foreach ([
             'body' => $bodyarray,
-            '*' => $astarray,
+            //'*' => $astarray,
```
Progi1984, line 104:
```suggestion
```

src/PhpWord/Writer/HTML/Element/Title.php
oleibman: Latest push has this change.

src/PhpWord/Writer/HTML/Part/Head.php
oleibman: Latest push has this change.

src/PhpWord/Writer/HTML/Element/Title.php
```diff
@@ -34,21 +38,39 @@ class Title extends AbstractElement
...
+        $paragraphStyle = null;
         if (is_string($text)) {
             $text = $this->parentWriter->escapeHTML($text);
         } else {
+            $paragraphStyle = $text->getParagraphStyle();
             $writer = new Container($this->parentWriter, $text);
             $text = $writer->write();
         }
+        $css = '';
+        $write1 = $write3 = '';
```
Progi1984:
```suggestion
        $cssFont = $cssParagraph = '';
```

src/PhpWord/Writer/HTML/Element/Title.php
```diff
@@ -34,21 +38,39 @@ class Title extends AbstractElement
...
+        $style = Style::getStyle('Heading_' . $this->element->getDepth());
+        if ($style !== null) {
+            $styleWriter = new Font($style);
+            $write1 = $styleWriter->write();
+        }
+        if (is_object($paragraphStyle)) {
+            $styleWriter = new Paragraph($paragraphStyle);
+            $write3 = $styleWriter->write();
+        }
+        $write2 = ($write1 !== '' && $write3 !== '') ? ' ' : '';
```
Progi1984:
```suggestion
        $cssSeparator = ($write1 !== '' && $write3 !== '') ? ' ' : '';
```

tests/PhpWordTests/SettingsTest.php
```diff
@@ -34,7 +34,7 @@ class SettingsTest extends TestCase
     /** @var string */
     private $defaultFontColor;
 
-    private $defaultFontSize;
+    //private $defaultFontSize;
```
Progi1984: Why remove it ?

composer.json
```diff
@@ -122,7 +122,7 @@
         "phpmd/phpmd": "^2.13",
         "phpstan/phpstan": "^0.12.88 || ^1.0.0 || ^2.0.0",
         "phpstan/phpstan-phpunit": "^1.0 || ^2.0",
-        "phpunit/phpunit": "^7.0 || ^8.0 || ^9.0 || ^10.0 || ^11.0",
+        "phpunit/phpunit": "^7.0 || ^8.0 || ^9.0 || ^10.0",
```
Progi1984: Can you revert this change ?

Progi1984 (commented):
@oleibman Can you rebase and fix my feedbacks ?

PR #2542, opened 2024-01-06
TemplateProcessor Persist File After Destruct

### LINKED ISSUE #2539: [DOC] TemplateProcessor behavior changed with 1.2 - temporary file gets deleted while still being used
### Describe the Bug

Hi,

thanks for your work on this library!

A recent change introduced a bug that I can workaround, but I am not sure if this is intended behavior, as it comes pretty unexpected. It was introduced in https://github.com/PHPOffice/PHPWord/pull/2475, which added `TemplateProcessor::__destruct()` in which the temporary file is deleted.

https://github.com/PHPOffice/PHPWord/blob/32a451da627528fe376d15e5348b7341b4c47f06/src/PhpWord/TemplateProcessor.php#L141-L153

This change basically disallows to use the temporary file while the process is still running, unless you keep the TemplateProcessor instance alive. Temporary files are, to my knowledge, always deleted by the engine after the scripts end, so this code seems to be superfluous.

### Steps to Reproduce

This is my code:
```php
$template = new TemplateProcessor('template.docx');
$tplFile = $template->save();
return $this->getFileResponse(new Stream($tplFile), 'export.docx');
```

This worked fine until I upgraded to 1.2

You save a Template and return the temporary filename, all from the context of a method that does uses a local instance of TemplateProcessor. As soon as your TemplateProcessor instance gets destructed (here when returning the response from the method) the temporary file now gets deleted.

I can work around the problem with this code:

```php
$tplFile = @tempnam(sys_get_temp_dir(), 'foo-bar-docx');
$template->saveAs($tplFile);
return $this->getFileResponse(new Stream($tplFile), 'export.docx');
```

So basically using my own temporary file. I feel that this change makes the save() method kind of useless.

### Expected Behavior

Do not delete the temporary file while I am still using it. 
Trust the user of your library: I called `save()`, I can/will handle the file myself. 
So my proposal is: remove the destruct method again.

### Current Behavior

The temporary file is deleted upon destruct of the TemplateProcessor instance, which is likely before the file is used (here sent to the user).

### Context

- PHP Version: 8.2
- PHPWord Version: 1.2

oleibman:
Fix #2539. Inadvertent break in TemplateProcessor behavior with #2475. Deleted temp file on destruct. It will now persist, restoring prior behavior, unless user specifies otherwise in constructor.

### Description

Fixes # (issue)

### Checklist:


src/PhpWord/TemplateProcessor.php
```diff
@@ -994,6 +982,12 @@ public function save()
         return $this->tempDocumentFilename;
     }
 
+    /** @return string */
+    public function getTempDocumentFilename()
```
Progi1984, line 986:
```suggestion
    public function getTempDocumentFilename(): string
```

Progi1984 (commented):
@oleibman Hi, could you update the changelog, please ?

src/PhpWord/TemplateProcessor.php
oleibman: Done. I wasn't sure if the old versions of Php that you still support allowed scalar typehints; I'm glad to see that they do.

oleibman:
@Progi1984 I updated change log for 1.2.0 (where break occurred) and 2.0.0 (where it will be fixed). If you have any suggested rewording, or anywhere else you want it made known, let me know. Apologies for creating the problem in the first place.

docs/changes/2.x/2.0.0.md
```diff
@@ -16,5 +17,6 @@
 - Bump phpmd/phpmd from 2.14.1 to 2.15.0 by [@dependabot](https://github.com/dependabot) in [#2538](https://github.com/PHPOffice/PHPWord/pull/2538)
 - Bump phpunit/phpunit from 9.6.14 to 9.6.15 by [@dependabot](https://github.com/dependabot) in [#2537](https://github.com/PHPOffice/PHPWord/pull/2537)
 - Bump symfony/process from 5.4.28 to 5.4.34 by [@dependabot](https://github.com/dependabot) in [#2536](https://github.com/PHPOffice/PHPWord/pull/2536)
+- Allow rgb() when converting Html [@oleibman](https://github.com/oleibman) fixing [#2508](https://github.com/PHPOffice/PHPWord/issues/2508) in [#2512](https://github.com/PHPOffice/PHPWord/pull/2512)
```
Progi1984, line 20:
```suggestion
```

Progi1984 (commented):
@oleibman Could you rebase and remove the merge commit ?

oleibman:
Github says I can't rebase due to merge conflicts. I will just re-create and submit a new PR.

oleibman:
Superseded by #2545.

PR #2545, opened 2024-01-08
Template Processor Persist File After Destruct

### LINKED ISSUE #2539: [DOC] TemplateProcessor behavior changed with 1.2 - temporary file gets deleted while still being used

oleibman:
### Description

Replace PR #2542.

Fix #2539. Inadvertent break in TemplateProcessor behavior after #2475. Deleted temp file on destruct. It will now persist after destructor.

### Checklist:

### DIFF
--- a/docs/changes/1.x/1.2.0.md
+++ b/docs/changes/1.x/1.2.0.md
@@ -65,3 +65,4 @@
 
 ### BC Breaks
 - Removed dependency `laminas/laminas-escaper`
+- *Unintended Break* TemplateProcessor Does Not Persist File After Destruct. [#2539](https://github.com/PHPOffice/PHPWord/issues/2539) To be fixed by [#2545](https://github.com/PHPOffice/PHPWord/pull/2545
--- a/docs/changes/2.x/2.0.0.md
+++ b/docs/changes/2.x/2.0.0.md
@@ -7,7 +7,7 @@
 ### Bug fixes
 
 - MsDoc Reader : Correct Font Size Calculation by [@oleibman](https://github.com/oleibman) fixing [#2526](https://github.com/PHPOffice/PHPWord/issues/2526) in [#2531](https://github.com/PHPOffice/PHPWord/pull/2531)
-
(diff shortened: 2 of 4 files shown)


Progi1984 (approved):

Progi1984:
Thanks @oleibman for your contribution

kevinpapst:
Thanks for tackling the issue 👍

Issue #1615, opened 2019-04-15
Issue templateProcessor setValue html display


minhnhsg:
Hi everyone,
I'm using PHPWord v0.16. I have a issue like this:
- Using templateProcessor and set setValue with HTML display (HTML code, with superscript, subscript, ...). I have searched for a solution for it many times, but no.
Anyone can help me with this issue?
Thank for your help, have a nice day!

yellow1912:
For simple HTML, you could use setComplexValue, something like this

$text = new TextRun();
Html::addHtml($text, 'your html here');
$templateProcessor->setComplexValue('replace', $text);

It doesn't work very well with complex, nested elements such as ordered list, I haven't found a good way to do it.

ddruganov:
i know this is a very old post but found a reasonably good solution that works:

```php
private function prepareHtml(string $value): AbstractElement
    {
        $content = html_entity_decode($value);

        $dom = new DOMDocument();
        try {
            $dom->loadXML($content);
        } catch (Throwable) {
            $content = "<p>$content</p>";
            $dom->loadXml($content);
        }
        $dom->documentElement->setAttribute('style', 'font-family: Tahoma; font-size: 12pt');
        $content = $dom->saveXML($dom->documentElement);

        $table = new Table(['cellMargin' => 0]);
        $table->addRow();
        $cell = $table->addCell();
        Html::addHtml($cell, $content);

        foreach ($cell->getElements() as $element) {
            if (method_exists($element, 'getParagraphStyle')) {
                $element->getParagraphStyle()->setIndent(0);
            }
        }

        return $table;
    }
```

@yellow1912 your solution might be valid but in my case throws a BadMethodCallException: "Cannot add TextRun in TextRun."

Issue #110, opened 2014-03-12
Ability to insert formatted html chunks to template.


egig:
A request: It'll be great if we can insert html chunk to doc template, which is urgently I need to do now. e.g:

``` php
$doc = $PHPWord->loadTemplate('Template.docx');

$doc->setValueFromHtml('SomePlaceholder', '<h1>some value</h1>');
```

Is there anyone know how to do this, just for now ? by hack or something ?
Thanks.

calita78:
Check this solution: http://stackoverflow.com/questions/27312321/solved-phpoffice-phpword-loadtemplate-not-rendering-word-as-it-should .... i've tested with success, using https://h2openxml.codeplex.com/

GarryOne:
The other answers, propose [H2OXML][1] which support just

> Bold, italic and underlined text 

> Bulled lists

from their docs and last update was in 2012, which wasn't a good option for me.

I did some research and found a pretty nice solution.
$var = 'Some text';
$xml = "<w:p><w:r><w:rPr><w:strike/></w:rPr><w:t>". $var."</w:t></w:r></w:p>";

$templateProcessor->setValue('param_1', $xml);

The above example, shows how would be a striked text. Instead of <w:strike/> you can use <w:i/> for italic or <w:b/> bold, and so on. Not sure if it works on all tags or not.

  [1]: https://h2openxml.codeplex.com/

pascaldls:
has this gone any where really need this function
and think this library need it as well

kindrosker:
You can add html to table cell, see example:

```
$value = '<b>TEST</b>';
$wordTable = new \PhpOffice\PhpWord\Element\Table();
$wordTable->addRow();
$cell = $wordTable->addCell();                                
 \PhpOffice\PhpWord\Shared\Html::addHtml($cell,$value);
                
$templateProcessor->setComplexBlock($block_id, $wordTable);
```

Warafux:
> You can add html to table cell, see example:
> 
> ```
> $value = '<b>TEST</b>';
> $wordTable = new \PhpOffice\PhpWord\Element\Table();
> $wordTable->addRow();
> $cell = $wordTable->addCell();                                
>  \PhpOffice\PhpWord\Shared\Html::addHtml($cell,$value);
>                 
> $templateProcessor->setComplexBlock($block_id, $wordTable);
> ```

Omg thank you very much. Never though of using a table... I've used everything but a table!
Thank you very much sir.

[...]

PR #2547, opened 2024-01-10
Add a method to TemplateProcessor for rendering HTML content.Include Image

Maybe-U:
### Description

Add a method to TemplateProcessor for rendering HTML content. support image 

Fixes  1
-  https://github.com/PHPOffice/PHPWord/issues/1615
-  https://github.com/PHPOffice/PHPWord/issues/110


Maybe-U:
> @Maybe-U You asked me to review. The code looks okay to me, but I'm really not all that familiar with TemplateProcessor, so it needs someone more knowledgeable to review. However, your test may have a problem. Does it actually attempt to read your external images? If the answer is no, there's no problem. If the answer is yes, I'll add some more comments.

Thank you very much for your response. My test cases only verified the generation of the DOCX file and did not check the actual content. However, in reality, it is possible to read external or local images, so my  answer is yes.
this is html content 
image
this is output docx file
image

Maybe-U:
> @Maybe-U You asked me to review. The code looks okay to me, but I'm really not all that familiar with TemplateProcessor, so it needs someone more knowledgeable to review. However, your test may have a problem. Does it actually attempt to read your external images? If the answer is no, there's no problem. If the answer is yes, I'll add some more comments.

oleibman:
My testing indicates that your test does attempt to load image https://t7.baidu.com/it/u=4198287529,2774471735&fm=193&f=GIF and will fail if it is not available; that is probably true of your second image as well. The problem is that external files tend to move or be deleted over time, and, when (not if) that happens, the unit test suite will fail. To avoid this problem, I think you might be able to use `AbstractWebServerEmbeddedTest` so that the requests are routed to localhost rather than an external site. Again, I'm not all that familiar with it, but there are a lot of tests that use it (e.g. `HtmlTest`).

Maybe-U:
> My testing indicates that your test does attempt to load image https://t7.baidu.com/it/u=4198287529,2774471735&fm=193&f=GIF and will fail if it is not available; that is probably true of your second image as well. The problem is that external files tend to move or be deleted over time, and, when (not if) that happens, the unit test suite will fail. To avoid this problem, I think you might be able to use `AbstractWebServerEmbeddedTest` so that the requests are routed to localhost rather than an external site. Again, I'm not all that familiar with it, but there are a lot of tests that use it (e.g. `HtmlTest`).

i am Optimize test cases and handle image loading failures in setHtmlBlock function please have a review  thanks！

src/PhpWord/TemplateProcessor.php
```diff
@@ -315,6 +318,123 @@ public function setComplexBlock($search, Element\AbstractElement $complexType):
...
+        $phpWord = new PhpWord();
+        $section = $phpWord->addSection();
+        //deal remote load Image
+        $pattern = '/<img[^>]+src\s*=\s*["\']([^"\']+)["\'][^>]*>/i';
+        preg_match_all($pattern, $htmlContent, $matches);
+        $imageSrcList = $matches[1];
+        if (!empty($imageSrcList)) {
+            foreach ($imageSrcList as $imageSrc) {
+                try {
+                    $content = file_get_contents($imageSrc);
```
oleibman, line 337: This section of the code is quite clever, but (a) I'm not really sure it's needed, and (b) I don't think you've coded it correctly. I would personally eliminate all the code after `addSection` and before `addHtml` and just let it fail later if the file isn't available; your changes to your test member are sufficient to address my original concern by making the files local rather than external. But, I can see that this checking might be perceived as a benefit, so let's discuss (b). `file_get_contents` does not normally throw an exception (I am not expert in the Php internals, so I suppose there might be some edge case where it throws); it normally just returns `false` if it fails and issues some warning messages describing the failure. So, I think what you want to do is:
```php
foreach ... {
    $content = @file_get_contents(...); // suppress warning messages
    if ($content === false) {
        $localImg = ...
        $htmlContent = ...
    }
}
```

githubguoyy (commented):
image

majeeed87:
<w:sectPr><w:pgSz w:orient="portrait" w:w="11905.511811023622" w:h="16837.79527559055"/><w:pgMar w:top="1440" w:right="1440" w:bottom="1440" w:left="1440" w:header="720" w:footer="720" w:gutter="0"/><w:cols w:num="1" w:space="720"/></w:sectPr>

Due to the section's properties are stored in the sectPr element above. The document style changed.
So, I use this line to delete the sectPr element before replaceXmlBlock.

$documentBodyStr = preg_replace('/<w:sectPr>.*?<\\/w:sectPr>/', '', $documentBodyStr);
//replace document.xml
$this->replaceXmlBlock($search, $documentBodyStr, 'w:p');

PR #2551, opened 2024-01-15
Fix 32-bit Problem in PasswordEncoder

### LINKED ISSUE #2550: 32-bit Error With PasswordEncoder
### Describe the Bug

PasswordEncoder is not 32-bit safe. PasswordEncoderTest will fail ("Implicit conversion from float 4294967295 to int loses precision") when tested with a recent enough version of 32-bit Php. This will cause the test to error out. It is easily fixed, and a PR will be available within an hour. AFAIK, this is the only 32-bit problem in the codebase.

### Steps to Reproduce

Run unit-test PasswordEncoderTest with a 32-bit version of Php

### Expected Behavior

Test should succeed.

### Current Behavior

What is the current behavior?
Test fails ("Implicit conversion from float 4294967295 to int loses precision").

### Context

- PHP Version: 8.3, 8.2
- PHPWord Version: any

oleibman:
### Description

Password encoder is not 32-bit safe.

Fixes #2550

### Checklist:


Progi1984 (commented):
@oleibman Approved

Could you update the changelog ?

oleibman:
@Progi1984 Change log is updated.

Progi1984 (approved):

PR #2554, opened 2024-01-18
Php7 Problem with TemplateProcessor Destructor

oleibman:
### Description

A particularly perplexing problem accidentally introduced by PR #2475. Problem does not arise for Php8, and does not arise for Php7 unit tests. But, running *not* under Phpunit auspices with Php7 can cause a warning message at destructor time if the `save` function has been used. A very artificial test is introduced to test this situation.

Fixes # 2548

### Checklist:

### DIFF
--- a/src/PhpWord/Shared/ZipArchive.php
+++ b/src/PhpWord/Shared/ZipArchive.php
@@ -20,6 +20,7 @@ namespace PhpOffice\PhpWord\Shared;
 use PclZip;
 use PhpOffice\PhpWord\Exception\Exception;
 use PhpOffice\PhpWord\Settings;
+use Throwable;
 
 /**
  * ZipArchive wrapper.
@@ -162,13 +163,16 @@ class ZipArchive
      * Close the active archive.
      *
      * @return bool
-     *
-     * @codeCoverageIgnore Can't find any test case. Uncomment when found.
      */
     public function close()
     {
         if (!$this->usePclzip) {
-            if ($this->zip->close() === false) {
+            try {
+                $result = @$this->zip->close();
+            } catch (Throwable $e) {
+                $result = false;
+            }
+            if ($result === false) {
                 throw new Exception("Could not close zip file {$this->filename}: ");
             }
         }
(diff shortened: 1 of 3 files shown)


src/PhpWord/Shared/ZipArchive.php
```diff
@@ -162,13 +163,16 @@ public function open($filename, $flags = null)
...
     public function close()
     {
         if (!$this->usePclzip) {
-            if ($this->zip->close() === false) {
+            try {
+                $result = @$this->zip->close();
+            } catch (Throwable $e) {
+                $result = false;
+            }
+            if ($result === false) {
```
Progi1984, line 175:
```suggestion
```

Progi1984: You can remove these lines.

Progi1984 (commented):
A small feedback

src/PhpWord/Shared/ZipArchive.php
oleibman: No, your suggestion changes the program logic. We should get `false` for the result of `close`. Php8 changed that, and I have filed a bug with Php about it. For Php7, we still get false, and we want to throw an exception (to avoid a compatibility break for Php7, which is what this change is all about). We also want to throw an exception when Php8 (erroneously) throws an exception for `close`. The way to handle both is to evaluate `$result` outside the try-catch, as I have currently coded it.

Maybe-U: > No, your suggestion changes the program logic. We should get `false` for the result of `close`. Php8 changed that, and I have filed a bug with Php about it. For Php7, we still get false, and we want to throw an exception (to avoid a compatibility break for Php7, which is what this change is all about). We also want to throw an exception when Php8 (erroneously) throws an exception for `close`. The way to handle both is to evaluate `$result` outside the try-catch, as I have currently coded it.

I would like to ask, how can I get Progi1984 to review my code?

oleibman: There is a big backlog, and I think he is being very careful and methodical in how he's trying to catch up. I'm sure he will get to your PR, perhaps not as quickly as you would like, but hopefully not too long.

bunglegrind:
In https://www.php.net/manual/en/ziparchive.close.php is stated:

> This method is automatically called at the end of the script. 

Maybe we don't need an explict call inside __destroy

PR #2555, opened 2024-01-18
Background and border color transparency support for textboxes when color is undefined

Chudy20007:
### Description

In project we use text boxes for imprints and side notes on a word document. It's currently not possible to use non-white background with border. As a simple fix I implement simple check if border or background color is set - if not then text box should be rendered with a transparent colors, else current behavior will be used.

Fixes # (issue)

### Checklist:


mzgs:
Thanks a lot. its woking.

Progi1984 (changes requested):
@Chudy20007 Could you add some unit tests and update the changelog, please ?

Progi1984 (approved):

Progi1984:
@Chudy20007 Thank you for your contribution 🥳

PR #2566, opened 2024-02-06
Alterd the function ensureUtf8Encoded to return only an empty string on null and false

SnipsMine:
### Description

I ran into a problem that nothing was printed when 0 was given as the value for function TemplateProcessor->setValue(). I tracked this issue down to the function ensureUtf8Encoded that checked if $subject was true, instead of is not null.

I changed the line to check if $subject is null  and only then return a empty string else the value is send to Text::toUTF8().

The test I wrote might be a bit overkill, but I hope it sufficient to test te fix

### Checklist:


Progi1984:
Superseeds by #2748

PR #2573, opened 2024-02-18
<w:lastRenderedPageBreak/> page break support

stanolacko:
### Description

Some Word2007 document are using <w:lastRenderedPageBreak/>for page break. When converting this document to PDF, it does not add page break in final PDF.

### DIFF
--- a/src/PhpWord/Reader/Word2007/AbstractPart.php
+++ b/src/PhpWord/Reader/Word2007/AbstractPart.php
@@ -199,7 +199,7 @@ abstract class AbstractPart
             $fontStyle = $this->readFontStyle($xmlReader, $domNode);
             $nodes = $xmlReader->getElements('w:r', $domNode);
             foreach ($nodes as $node) {
-                if ($xmlReader->elementExists('w:lastRenderedPageBreak',$node)) {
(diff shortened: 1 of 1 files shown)


src/PhpWord/Reader/Word2007/AbstractPart.php
```diff
@@ -199,6 +199,9 @@ protected function readParagraph(XMLReader $xmlReader, DOMElement $domNode, $par
             $fontStyle = $this->readFontStyle($xmlReader, $domNode);
             $nodes = $xmlReader->getElements('w:r', $domNode);
             foreach ($nodes as $node) {
+                if ($xmlReader->elementExists('w:lastRenderedPageBreak',$node)) {
```
Progi1984, line 202:
```suggestion
                if ($xmlReader->elementExists('w:lastRenderedPageBreak', $node)) {
```

Progi1984 (commented):
Could you add a Line in the Changelog and a unit test ?

stanolacko:
What kind of test is requered? For example simple text, with <w:lastRenderedPageBreak/> which generated 2 pages document?

Progi1984:
Superseeded by #2662

PR #2600, opened 2024-04-02
[BUGFIX] Remove invalid tool call from composer `check` script

sbuerk:
With commit [1] development dependencies have been removed
from the `composer.json`, but not from the `composer.lock`.

This indicates a inproper use of `composer` and should be
taken care in review of pull-requests. A good practice here
is to etablish a rule that composer commands needs to be
added to the commit message and the pull-request.

Other removed dependency and downgraded dependencies have
been fixed or readded meanwhile, still missing the `phpcs`
tool for the `check composer script`.

The `CONTRIBUTION.md` states to execute `composer check`
before creating a pull-request, which is literally broken
due to the missing dependency.

This change removes the php sniffer call from the check
script. Additionally, the composer scripts are enhanced
to use the same php binary used to invoke composer itself
to mitigate issues using the wrong php version on systems
with multiple php version binaries.

Further consideration should be to use the check command in
the Github Action workflow to have a border if development
toolchain is gonna be broken with a change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,20 +36,19 @@
     ],
     "scripts": {
         "test": [
-            "phpunit --color=always"
+            "@php vendor/bin/phpunit --color=always"
         ],
         "test-no-coverage": [
-            "phpunit --color=always --no-coverage"
+            "@php vendor/bin/phpunit --color=always --no-coverage"
         ],
         "check": [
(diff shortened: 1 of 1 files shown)


Progi1984 (approved):

Progi1984:
Thanks @sbuerk for your contribution

PR #2604, opened 2024-04-04
FIX: StrikeThrough property in word2007

noec764:
I had a problem with strike through in Word2007.
If a strikethrough style was added to a text element it never appeared in the resulting document.
Indeed, the XML generated was as follows
```
<w:r>
    <w:rPr>
        <w:strike w:val="1"/>
        <w:dstrike w:val="0"/>
     </w:rPr>
     <w:t xml:space="preserve">some text</w:t>
</w:r>
```
The statement `<w:dstrike w:val="0"/>` should not be present here
I figure out that the test  `$style->isDoubleStrikethrough() !== null` is true if `$style->isDoubleStrikethrough() === false`
So as a text cannot be strike and doublestrike at the same time, word2007 ignores it.


Progi1984:
@noec764 Hi, Could you update the changelog and add some unit tests, please ?

Progi1984:
Superseeded by #2661

PR #2625, opened 2024-06-18
Fix unknown method when not using PCLZip library

glafarge:
Fix unknown method when not using PCLZip library


Progi1984:
@glafarge Hi, could you rebase your PR, update the changelog, and update `phpstan-baseline.neon` to ignore the error ?

PR #2637, opened 2024-07-28
Added PHP end tag to example in readme

lukmi15:
Is there any reason the example in the readme does not have a PHP end tag? If not, then here's a PR.


sbuerk:
Usually end php tag should be omited in pure code files. End tags are only used in mixed markup (html) and php files aka template files to close blogs before switching to files.

Using end tags is discouraged since decades in PHP. Cannot speak for this project, but I would not add it to example snippets.

lukmi15:
Learned something new, thanks

PR #2661, opened 2024-08-27
Word2007 Writer : Fixed StrikeThrough property

### LINKED ISSUE #1693: Strikethrough not working in v0.16.0
### Describe the Bug

Strikethrough is no longer working in version 0.16.0, it is working in version 0.15.0

### Steps to Reproduce

```php
<?php

$section->addText('Hello World!', array('strikethrough' => true));

$objWriter->save('/tmp/test.docx');
```
### Expected Behavior

In the above example 'Hello World!' should have the ~strikethrough~ formatting applied.  

Using "composer require phpoffice/phpword" to install the PHPWord results in strikethrough not being applied.  However using "composer require phpoffice/phpword:0.15.0" to install the PHPWord does apply strikethrough.

### Context
Ubuntu: 19.04
PHP Version: 7.2.19
LibreOffice: 6.2.5.2

### LINKED ISSUE #1722: Strikethrough font style in Word2007 does not work
It's not possible to use strike font style with Word2007 writer because it uses strict comparison. Methods `isStrikethrough` and `isDoubleStrikethrough` can return `true`, `false` or `null`, `false !== null`.

src/PhpWord/Writer/Word2007/Style/Font.php
```php
// Strikethrough, double strikethrough
$xmlWriter->writeElementIf($style->isStrikethrough() !== null, 'w:strike', 'w:val', $this->writeOnOf($style->isStrikethrough()));
$xmlWriter->writeElementIf($style->isDoubleStrikethrough() !== null, 'w:dstrike', 'w:val', $this->writeOnOf($style->isDoubleStrikethrough()));
```

You cannot use elements `w:strike` and `w:dstrike` in same time. When you set strike style to true eg. with `setStrikethrough` then it calls method `setPairedVal` that should disable double strike style.

src/PhpWord/Style/Font.php
```php
    /**
     * Set strikethrough
     *
     * @param bool $value
     * @return self
     */
    public function setStrikethrough($value = true)
    {
        return $this->setPairedVal($this->strikethrough, $this->doubleStrikethrough, $value);
    }
```

src/PhpWord/Style/AbstractStyle.php
```php
    /**
     * Set $property value and set $pairProperty = false when $value = true
     *
     * @param bool &$property
     * @param bool &$pairProperty
     * @param bool $value
     * @return self
     */
    protected function setPairedVal(&$property, &$pairProperty, $value)
    {
        $property = $this->setBoolVal($value, $property);
        if ($value === true) {
            $pairProperty = false;
        }
        return $this;
    }
```

In the writer the method `$style->isDoubleStrikethrough() !== null` evaluates as `true` even when this method returns `false`. Please note that method `setPairedVal` sets this value to `false`.

Progi1984:
### Description

Word2007 Writer : Fixed StrikeThrough property

Superseeds #2604 by @noec764

Fixes #1693
Fixes #1722

> If a strikethrough style was added to a text element it never appeared in the resulting document.
> Indeed, the XML generated was as follows
> ```
><w:r>
>    <w:rPr>
>        <w:strike w:val="1"/>
>        <w:dstrike w:val="0"/>
>     </w:rPr>
>     <w:t xml:space="preserve">some text</w:t>
></w:r>
>```
> The statement `<w:dstrike w:val="0"/>` should not be present here
> I figure out that the test  `$style->isDoubleStrikethrough() !== null` is true if `$style->isDoubleStrikethrough() === false`
> So as a text cannot be strike and doublestrike at the same time, word2007 ignores it.

### Checklist:

- [x] My CI is :green_circle:
- [x] I have updated the [changelog](https://github.com/PHPOffice/PHPWord/blob/master/docs/changes/2.x/2.0.0.md)

### DIFF
--- a/docs/changes/1.x/1.3.0.md
+++ b/docs/changes/1.x/1.3.0.md
@@ -28,6 +28,7 @@
 - Word2007 Writer : Fix first footnote appearing as separator [#2634](https://github.com/PHPOffice/PHPWord/issues/2634) by [@jacksleight](https://github.com/jacksleight) in [#2635](https://github.com/PHPOffice/PHPWord/pull/2635)
 - Template Processor : Fixed images with transparent backgrounds displaying a white background by [@ElwynVdb](https://github.com/ElwynVdb) in [#2638](https://github.com/PHPOffice/PHPWord/pull/2638)
 - HTML Writer : Fixed rowspan for tables by [@andomiell](https://github.com/andomiell) in [#2659](https://github.com/PHPOffice/PHPWord/pull/2659)
+- Word2007 Writer : Fixed StrikeThrough property by [@noec764](https://github.com/noec764) fixing [#1722](https://github.com/PHPOffice/PHPWord/issues/1722) & [#1693](https://github.com/PHPOffice/PHPWord/issues/1693) in [#2661](https://github.com/PHPOffice/PHPWord/pull/2661)
 
 ### Miscellaneous
 
--- a/src/PhpWord/Style/Font.php
+++ b/src/PhpWord/Style/Font.php
@@ -565,10 +565,8 @@ class Font extends AbstractStyle
 
     /**
      * Get strikethrough.
-     *
-     * @return bool
      */
-    public function isStrikethrough()
+    public function isStrikethrough(): ?bool
     {
         return $this->strikethrough;
     }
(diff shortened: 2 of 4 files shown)


Progi1984:
Thanks @noec764 for your contribution :)

PR #2685, opened 2024-09-22
Fix that style objects cannot be passed to element constructors

hazington:
### Description

I generally prefer to use style objects over style arrays in elements. But, if you pass a `\PhpOffice\PhpWord\Style\Cell` style element to the `\PhpOffice\PhpWord\Element\Cell` constructor then the `$returnObject: true` argument of the `setNewStyle()` method in the constructor completely ignores the passed style element, because the `setNewStyle()` method only checks if `$styleValue` is an array and returns `$styleObject` as empty style object otherwise. Imho, if `$styleValue` is a style object with the same type as `$styleObject` then the `$returnObject: true` should also be "ignored" and the passed style object should then be returned instead of the empty base style object.

Fixes # (issue)

### Checklist:

- [x] My CI is :green_circle:
- [x] ~I have updated the [documentation](https://github.com/PHPOffice/PHPWord/tree/master/docs) to describe the changes~ *(not applicable, bugfix did not change the behavior)*
- [x] I have updated the [changelog](https://github.com/PHPOffice/PHPWord/blob/master/docs/changes/1.x/1.4.0.md)


hazington:
It took me a while to find the cause of the phpstan errors. Never imagined that there would be such a long list of ignored errors.

hazington:
@Progi1984 Is there a chance to look over this pull request? I see that PhpWord transforms to a strict typed library and thus elements should accept style objects. This is also necessary for a better html to word implementation. At the moment you can't pass style objects since they are ignored.

hazington:
@Progi1984 ?

PR #2741, opened 2025-02-04
TemplateProcessor: Added a method to replace multiple XML blocks

sr-44:
### Description

Often you will need to replace variables with a new XML block. Currently, there is already a `replaceXmlBlock` method that replaces only the first occurrence of a variable. To replace all of them, you will need to call this method again. This makes the code unreadable, and sometimes there is a chance that the number of these variables is determined dynamically (in my case it is). I have added a new public method `replaceMultipleXmlBlocks`, which uses the new protected method `findAllContainingXmlBlockForMacro` and solves this problem. I would be glad if my contribution is accepted!

### Checklist:

- [x] My CI is :green_circle:
- [ ] I have updated the [changelog](https://github.com/PHPOffice/PHPWord/blob/master/docs/changes/1.x/1.4.0.md)

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -1354,10 +1354,39 @@ class TemplateProcessor
     }
 
     /**
+     * Replace an XML block surrounding a macro with a new block.
+     *
+     * @param string $macro Name of macro
+     * @param string $block New block content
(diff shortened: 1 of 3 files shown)


sr-44:
@Progi1984 can you please review this PR?

PR #2700, opened 2024-11-21
Add default font color for Word (.docx)

Collie-IT:
Adds the abillity to add a default font color for generated .docx.

### Description

Adds the abillity to add a default font color for generated .docx.

Fixes # (issue)

### Checklist:

- [x] My CI is :green_circle:
- [x] I have covered by unit tests my new code (check build/coverage for coverage report)
- [x] I have updated the [documentation](https://github.com/PHPOffice/PHPWord/tree/master/docs) to describe the changes
- [x] I have updated the [changelog](https://github.com/PHPOffice/PHPWord/blob/master/docs/changes/1.x/1.4.0.md)

PR #2745, opened 2025-02-09
test the reader with the writer

MichaelPFrey:
### Description

I think that currently, there are only few good tests for the reader. (see for e.g. #2700 where the reader and writer where modified, but only the writer is actually covered by a test)

Creating and adding artifacts (docx and odt) is not ideal, as they are hard to verify and will contain random other artifacts.

By dynamically creating the test-file using the writer and reading it back in with the reader, we can simply test the reader against the writer.

In general, a WYSWYG word processor is assumed to read and save files without loosing anything. For a project like PHPOffice, the focus is stronger on the writer then the reader. With this setup, it is a lot easier to work on the reader, as a test driven development purely in PHP gets enabled - the need to use a word processor for verification gets significantly reduced.

There is also simply charm in self testing, as a simple to read test makes a literal "end to end" test, covering a lot of complex code with a few lines of simple code.

Fixes # (issue)

### Checklist:

- [x] My CI is :green_circle:
- [x] I have updated the [changelog](https://github.com/PHPOffice/PHPWord/blob/master/docs/changes/1.x/1.4.0.md)

This are literal test. They do not effect the production code. So the checklist does not directly apply.


Progi1984:
@MichaelPFrey Could you update the changelog, please ?

Progi1984 (approved):

Progi1984:
@MichaelPFrey Thank you for your contribution 🥳

PR #2757, opened 2025-02-23
Fix text extraction from PhpWord TextRun elements

### LINKED ISSUE #1625: TextRun in addTitle causes TOC error.
This is:
- [* ] a bug report

So I've been having an issue for a while regarding tabs in titles. Tabs do not work at all, looking deeper into the code I have seen that titles actually support TextRuns. Using a TextRun in the title causes the tab issue to go away. However, the docx file now errors on open (although the content is all there) and TOC has issues (Text missing).

### How to Reproduce

```php
<?php

$titleText = new TextRun();
$titleText->addText("Hello \t");
$titleText->addText('Some other text');
$this->sceneSection->addTitle($titleText, 1);
```

chalda-pnuzig:
### Description

Previously, `$title->getText()` was assumed to always return a string, but in some cases, it returns a `PhpOffice\PhpWord\Element\TextRun`. This commit adds a check to determine if the returned value is a `TextRun` and properly extracts text from its child elements.

Fixes #1625

### DIFF
--- a/src/PhpWord/Writer/Word2007/Element/TOC.php
+++ b/src/PhpWord/Writer/Word2007/Element/TOC.php
@@ -96,7 +96,17 @@ class TOC extends AbstractElement
         $xmlWriter->startElement('w:t');
 
         $titleText = $title->getText();
-        $this->writeText(is_string($titleText) ? $titleText : '');
+        if (get_class($titleText) === 'PhpOffice\PhpWord\Element\TextRun') {
+            $textRunElements = $title->getText()->getElements();
+            $uploadedText = '';
+            foreach ($textRunElements as $textRunElement) {
+                $uploadedText .= $textRunElement->getText();
+                $uploadedText .= ' ';
+            }
+            $this->writeText($uploadedText);
+        } else {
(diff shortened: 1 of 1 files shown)


src/PhpWord/Writer/Word2007/Element/TOC.php
```diff
@@ -96,7 +96,17 @@ private function writeTitle(XMLWriter $xmlWriter, TOCElement $element, Title $ti
...
+        if (get_class($titleText) === 'PhpOffice\PhpWord\Element\TextRun') {
+            $textRunElements = $title->getText()->getElements();
+            $uploadedText = '';
+            foreach ($textRunElements as $textRunElement) {
+                $uploadedText .= $textRunElement->getText();
+                $uploadedText .= ' ';
+            }
+            $this->writeText($uploadedText);
+        } else {
+            $this->writeText((is_string($titleText) ? $titleText : '');
```
Progi1984, line 108:
```suggestion
            $this->writeText(is_string($titleText) ? $titleText : '');
```

PR #2814, opened 2025-08-31
Various RTF Enhancements

rasamassen:
### Description

Adds RTF features, including many missing font settings, missing tab stop settings, vertical align, and fixes an error with textBreak inside a textRun (line return, but not new paragraph).

### More detailed information
TextBreak - returns \line when not a paragraph return.
Font - adds support for various underlines, smallcaps, allcaps, hidden, scale, spacing, kerning, position, and noproof. Support for color, fgcolor, bgcolor, shading, and lang not implemented, but placeholder code written for future work.
Section - adds support for vertical align.
Tab - adds support for tabLeader, as well as support for tabBar.

### DIFF
--- a/src/PhpWord/Writer/RTF/Element/TextBreak.php
+++ b/src/PhpWord/Writer/RTF/Element/TextBreak.php
@@ -36,6 +36,10 @@ class TextBreak extends AbstractElement
         $parentWriter = $this->parentWriter;
         $parentWriter->setLastParagraphStyle();
 
-        return '\pard\par' . PHP_EOL;
+        if ($this->withoutP) {
+            return '\line' . PHP_EOL;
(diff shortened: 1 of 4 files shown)


rasamassen:
It looks like most of the tests fail because the update no longer sets \cf unnecessarily.
Before:         $content .= '\cf' . $this->colorIndex;
Now:        $content .= $this->getValueIf($style->getColor() !== null, '\cf' . $this->colorIndex);
^ it doesn't matter either way because non-default colors currently aren't being added to the style sheet properly in header.php

A few of the tests fail because font is now returning the RTF tags in a slightly different order.

I've done the part that matters for what I needed on my site, so if someone else wants to clean up the tests, go for it.

PR #2818, opened 2025-09-01
RTF Writer: Fix Header.php registerFont() not registering fonts/colors in subElements (such as textRuns)

rasamassen:
registerFont() registered fonts and colors from sections, but not from subparts of sections, such as textruns.

### Description

Now the header registers every font and color in a textRun. This is important if the fonts and colors are actually going to be used.

Fixes # (issue)

### Checklist:

- [x] My CI is :green_circle:


rasamassen:
I don't understand why this is failing the Static Analysis. I'm not sure what codebase you're using for that, but it sounds a lot like the issue described here: https://github.com/phpstan/phpstan/issues/4579. I don't think this should be failing Static Analysis.

PR #2826, opened 2025-09-13
RTF Writer: PreserveText - Create

rasamassen:
### Description

Create PreserveText

### Checklist:

- [x] My CI is :green_circle:


rasamassen:
The description of preserveText in (Samples)[https://github.com/PHPOffice/PHPWord/blob/master/docs/usage/elements/preservetext.md] suggests that preserveText can be used to add Fields. This feels odd since there is already a field element? My current commit adds fields as described.

But what is the point of preserveText? preserveText sounds like it's supposed to be used for adding things like whitespace and special characters (like braces and slashes), but text already does that just fine. Instead of "preserving" the text, it currently adds fields.

So though I've built this like described, I'm confused what the purpose of the element is?

oleibman:
You are correct that PreserveText isn't absolutely required, but it can make things a lot simpler.
```php
$footer->addPreserveText('Page {PAGE} of {NUMPAGES}');
```
is much easier than adding a TextRun consisting of Text/Field/Text/Field.

PR #2828, opened 2025-09-16
Fix: prevent null being passed to htmlspecialchars() in AbstractPart.php

dafydd-rhys:
### Description

This PR fixes deprecation warnings in PHP 8.1+ caused by `htmlspecialchars()` receiving `null`.  
The affected code paths occur when XML nodes or text content may be missing.  

Fix applied by using the null coalescing operator (`?? ''`) to ensure a string is always passed.

"DEPRECATE: vendor/phpoffice/phpword/src/PhpWord/Reader/Word2007/AbstractPart.php line 300 user unknown -- htmlspecialchars(): Passing null to parameter [0000001](https://mantis.hoowla.com/view.php?id=1) ($string) of type string is deprecated"

### Checklist:

- [x] My CI is :green_circle:
- [x] I have verified the fix removes the PHP 8.1 deprecation warning

### DIFF
--- a/docs/changes/1.x/1.5.0.md
+++ b/docs/changes/1.x/1.5.0.md
@@ -7,6 +7,7 @@
 ### Bug fixes
 
 - Set writeAttribute return type by [@radarhere](https://github.com/radarhere) fixing [#2204](https://github.com/PHPOffice/PHPWord/issues/2204) in [#2776](https://github.com/PHPOffice/PHPWord/pull/2776)
+- Prevent `htmlspecialchars()` deprecation warnings by defaulting `null` values to empty strings in `AbstractPart` by [@dafydd-rhys](https://github.com/dafydd-rhys) fixing [#2829](https://github.com/PHPOffice/PHPWord/issues/2829) in [#2828](https://github.com/PHPOffice/PHPWord/pull/2828)
 
 ### Miscellaneous
 
--- a/src/PhpWord/Reader/Word2007/AbstractPart.php
+++ b/src/PhpWord/Reader/Word2007/AbstractPart.php
@@ -302,7 +302,7 @@ abstract class AbstractPart
             $nodes = $xmlReader->getElements('w:r|w:hyperlink', $domNode);
             $hasRubyElement = $xmlReader->elementExists('w:r/w:ruby', $domNode);
             if ($nodes->length === 1 && !$hasRubyElement) {
-                $textContent = htmlspecialchars($xmlReader->getValue('w:t', $nodes->item(0)), ENT_QUOTES, 'UTF-8');
+                $textContent = htmlspecialchars($xmlReader->getValue('w:t', $nodes->item(0)) ?? '', ENT_QUOTES, 'UTF-8');
             } else {
                 $textContent = new TextRun($paragraphStyle);
                 foreach ($nodes as $node) {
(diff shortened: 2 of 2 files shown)


Bosun18:
Had a look at this branch locally because the red checks looked odd for a three-line change.

The six PHP Static Analysis failures aren't coming from this PR. I ran phpstan on the branch tip and on its base commit `0ab0b4940` with the patch reverted, and the two outputs are byte-identical, same 180 errors either way. The one the annotations point at, `Offset bool|int|string might not exist on list<string|null>` on line 372, was silenced upstream in `5579bd257` (#2890, 16 Jul) with a `@phpstan-ignore` annotation, and this branch predates that by about a year. Rebased onto current master it gives `[OK] No errors`. As far as I can tell 7.2 and 7.3 stayed green only because phpstan 2.x needs PHP 7.4+, so composer resolves 1.x there.

PHPUnit 7.3 I can't say much about, the run logs are past GitHub's 90-day retention now so all that's left is metadata. It passed on `e9218c3f3` and failed on `2bfe4de55`, and the only difference between those two commits is one markdown line in the changelog. That job is also the only one running with xdebug and clover coverage. On green master runs it takes about 26 minutes, so the 22 minutes before it failed may well be the runner rather than the tests.

The other thing I ran into is that `table.docx` attached to #2829 doesn't actually reproduce the deprecation. I instrumented all six `htmlspecialchars()` calls in `AbstractPart` and read that file both on master and on the 1.4.0 tag: the calls fire three times and get null zero times. Its empty table cells are `<w:p>` with no `<w:r>` inside, so they go down the `addTextBreak` path and never reach `htmlspecialchars`. What does reproduce is a heading whose only run carries no `w:t` — a heading that is just a line break, or just an image:

```php
public function testReadHeadingWithoutText(): void
{
    $documentXml = '<w:p>
        <w:pPr>
            <w:pStyle w:val="Heading1" />
        </w:pPr>
        <w:r>
            <w:br />
        </w:r>
    </w:p>';

    $phpWord = $this->getDocumentFromString(['document' => $documentXml]);

    $elements = $phpWord->getSection(0)->getElements();
    self::assertInstanceOf('PhpOffice\PhpWord\Element\Title', $elements[0]);
    /** @var \PhpOffice\PhpWord\Element\Title $title */
    $title = $elements[0];
    self::assertSame('', $title->getText());
}
```

On master that errors with `htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated ... AbstractPart.php 305`, and with your change it passes. No special error handling needed in the test itself, `tests/bootstrap.php` already turns `E_DEPRECATED` into an exception.

One observation about the other two lines, and I might be misreading it. Across the full suite only line 305 ever receives null. `$fallback->nodeValue` is read after the `if ($fallbackElements->length)` check, so it's a `DOMElement` and its `nodeValue` is a string, and the `?string` looks like the stub type rather than something reachable at runtime. And `getValue('.', $node)` resolves the context node itself, so `getElements()` comes back with length 1. Both are harmless as defensive code, just not covered by anything, so it's worth deciding whether they belong in the same change.

The branch has also picked up a conflict since master moved, though it's only the neighbouring lines in `docs/changes/1.x/1.5.0.md`; `AbstractPart.php` still merges clean.

I can open the test as a PR against your branch if you'd rather keep it in one place, or separately, whichever suits. The rebase is what unblocks the checks either way.

Progi1984 (approved):

Progi1984:
Could you rebase and fix conflicts ?

Bosun18:
@dafydd-rhys if you're short on time, I have your commit rebased onto current master locally. The conflict is only the neighbouring lines in `docs/changes/1.x/1.5.0.md`, `AbstractPart.php` applies clean, and phpstan on the rebased branch comes back with `[OK] No errors`, so the six static analysis failures go away with the rebase.

I can open that as a PR with the authorship on your commit left as it is, plus the failing test from my comment above, if that saves you the round trip. Either way suits me, it's your patch.

Progi1984:
Fixed in #2932

PR #2873, opened 2026-05-04
Word2007 Writer: Image alt text support

### LINKED ISSUE #2630: Add the ability to add alt text to images
### Describe the problem

Currently, there appears to be no means to add alt text to images in Word2007 Writer. Alt text is important because it is displayed in lieu of the image if the image fails to load for some reason. It is also an important accessibility feature, since it allows screen readers to describe the content of an image if the image is not visible to a user.

### Describe the expected behavior

The addImage usage should change from:
$section->addImage($src, [$style]);
to:
$section->addImage($src, [$style], string $altText);

I have code that adds this capability, but I need to know if the $name variable is used with images, so that I can avoid conflicts in the order of Image properties.
In file src/PhpWord/Element/Image.php, line 149 shows the following:
public function __construct($source, $style = null, $watermark = false, $name = null)

The $name variable is not addressed in any documentation that I found, but it does appear on line 73 of tests/PhpWordTests/Element/ImageTest.php. If it is not used, then I would like to change the above constructor as follows, since it will provide a simpler call to add an image that is not a watermark:
public function __construct($source, $style = null, $altText = null, $watermark = false, $name = null)

If $name is used, then I will put the new $altText property at the end:
public function __construct($source, $style = null, $watermark = false, $name = null, $altText = null)

### Priority

- [ ] I want to crowdfund the feature (with [@algora-io](https://docs.algora.io/bounties/overview)) and fund a community developer.
- [ ] I want to pay the feature and fund a maintainer for that. (Contact @Progi1984)

jwoodhead:
### Description

Added support for specifying alt text (`descr`) on images. 

Fixes #2630.

### Checklist:

- [X] My CI is :green_circle:
- [X] I have covered by unit tests my new code (check build/coverage for coverage report)
- [X] I have updated the [documentation](https://github.com/PHPOffice/PHPWord/tree/master/docs) to describe the changes
- [X] I have updated the [changelog](https://github.com/PHPOffice/PHPWord/blob/master/docs/changes/1.x/1.5.0.md)

### DIFF
--- a/src/PhpWord/Element/Image.php
+++ b/src/PhpWord/Element/Image.php
@@ -69,9 +69,16 @@ class Image extends AbstractElement
     /**
      * Name of image.
      *
-     * @var string
+     * @var null|string
      */
     private $name;
+
+    /**
+     * Image alt text.
+     *
+     * @var null|string
+     */
+    private $altText;
 
     /**
      * Image type.
@@ -145,14 +152,16 @@ class Image extends AbstractElement
      * @param string $source
      * @param mixed $style
      * @param bool $watermark
-     * @param string $name
+     * @param null|string $name
+     * @param null|string $altText
      */
-    public function __construct($source, $style = null, $watermark = false, $name = null)
+    public function __construct($source, $style = null, $watermark = false, $name = null, $altText = null)
     {
         $this->source = $source;
         $this->style = $this->setNewStyle(new ImageStyle(), $style, true);
         $this->setIsWatermark($watermark);
         $this->setName($name);
+        $this->setAltText($altText);
 
         $this->checkImage();
     }
@@ -190,11 +199,31 @@ class Image extends AbstractElement
     /**
      * Sets the image name.
      *
-     * @param string $value
+     * @param null|string $value
      */
     public function setName($value): void
     {
         $this->name = $value;
+    }
+
+    /**
+     * Get image alt text.
+     *
+     * @return null|string
+     */
+    public function getAltText()
+    {
+        return $this->altText;
+    }
+
+    /**
+     * Sets the image alt text.
+     *
+     * @param null|string $value
+     */
+    public function setAltText($value): void
+    {
+        $this->altText = $value;
     }
 
     /**
--- a/src/PhpWord/Element/Image.php
+++ b/src/PhpWord/Element/Image.php
@@ -211,7 +211,7 @@ class Image extends AbstractElement
      *
      * @return null|string
      */
-    public function getAltText()
+    public function getAltText(): ?string 
     {
         return $this->altText;
     }
@@ -221,7 +221,7 @@ class Image extends AbstractElement
      *
      * @param null|string $value
      */
-    public function setAltText($value): void
+    public function setAltText(?string $value): void
     {
         $this->altText = $value;
     }
--- a/src/PhpWord/Writer/HTML/Element/Image.php
+++ b/src/PhpWord/Writer/HTML/Element/Image.php
@@ -45,8 +45,15 @@ class Image extends Text
             $style = $styleWriter->write();
             $imageData = 'data:' . $this->element->getImageType() . ';base64,' . $imageData;
 
+            $altText = $this->element->getAltText();
+            if ($altText) {
+                $altAttribute = " alt=\"{$altText}\"";
+            } else {
+                $altAttribute = '';
+            }
+
             $content .= $this->writeOpening();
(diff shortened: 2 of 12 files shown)


src/PhpWord/Element/Image.php
```diff
@@ -190,13 +199,33 @@ public function getSourceType()
...
     {
         $this->name = $value;
     }
 
+    /**
+     * Get image alt text.
+     *
+     * @return null|string
+     */
+    public function getAltText()
```
Progi1984:
```suggestion
    public function getAltText(): ?string 
```

src/PhpWord/Element/Image.php
```diff
@@ -190,13 +199,33 @@ public function getSourceType()
...
+    {
+        return $this->altText;
+    }
+
+    /**
+     * Sets the image alt text.
+     *
+     * @param null|string $value
+     */
+    public function setAltText($value): void
```
Progi1984:
```suggestion
    public function setAltText(?string $value): void
```

src/PhpWord/Writer/HTML/Element/Image.php
```diff
@@ -45,8 +45,15 @@ public function write()
             $style = $styleWriter->write();
             $imageData = 'data:' . $this->element->getImageType() . ';base64,' . $imageData;
 
+            $altText = $this->element->getAltText();
+            if ($altText) {
+                $altAttribute = " alt=\"{$altText}\"";
+            } else {
+                $altAttribute = '';
+            }
```
Progi1984, line 52:
```suggestion
            $altAttribute = '';
            if ($altText) {
                $altAttribute = " alt=\"{$altText}\"";
            }
```

tests/PhpWordTests/Reader/Word2007/ElementTest.php
```diff
@@ -338,7 +338,7 @@ public function testReadDrawing(): void
                             <a:graphicData uri="http://schemas.openxmlformats.org/drawingml/2006/picture">
                                 <pic:pic xmlns:pic="http://schemas.openxmlformats.org/drawingml/2006/picture">
                                     <pic:nvPicPr>
-                                        <pic:cNvPr id="1" name="file_name.jpg"/>
+                                        <pic:cNvPr id="1" name="file_name.jpg" descr="file description" />
```
Progi1984, line 341: Could you add an assert for checking the alt ?

src/PhpWord/Element/Image.php
```diff
@@ -145,14 +152,16 @@ class Image extends AbstractElement
      * @param string $source
      * @param mixed $style
      * @param bool $watermark
-     * @param string $name
+     * @param null|string $name
+     * @param null|string $altText
      */
-    public function __construct($source, $style = null, $watermark = false, $name = null)
+    public function __construct($source, $style = null, $watermark = false, $name = null, $altText = null)
```
Progi1984, line 158: Could you add unit tests for this part ?

src/PhpWord/Writer/Word2007/Element/Image.php
```diff
@@ -81,6 +81,10 @@ private function writeImage(XMLWriter $xmlWriter, ImageElement $element): void
         $xmlWriter->writeAttribute('type', '#_x0000_t75');
         $xmlWriter->writeAttribute('stroked', 'f');
 
+        if ($element->getAltText() != null) {
```
Progi1984, line 84: Could you add unit tests for this part ?

tests/PhpWordTests/Reader/Word2007/ElementTest.php
jwoodhead: This will require some enhancements to the `getDocumentFromString` method to support generating the associated `_rel` file.

The [`readRunChild`](https://github.com/PHPOffice/PHPWord/blob/0ab0b4940bc52c7183e82ab2fd55324607037a73/src/PhpWord/Reader/Word2007/AbstractPart.php#L536) method checks that the relationship exists for an image before adding it to the text run.

I did some logging during development to make sure it was able to retrieve the `name`, `altText`, and `embedId` from the XML.

My initial thoughts on how to support this is to add an optional `$relXmls` parameter to the `getDocumentFromString` method that uses similar logic to `$partXmls` to add relationship XML files to the `temp.docx` file.

Progi1984:
@jwoodhead Can you update the changelog file, please ?

Progi1984 (approved):

Progi1984:
Thanks @jwoodhead for your contribution.

PR #2878, opened 2026-05-20
Fix #2876: Handle large images in Mpdf writer without crashing pcre.backtrack_limit

### LINKED ISSUE #2876: Mpdf export can exceed pcre.backtrack_limit

idan57570idan-svg:
@oleibman @mrinaldidfs — Thank you both for the thorough feedback.

You were absolutely right, @oleibman: the previous PR direction removing the loop entirely was wrong and would have made things worse. I've completely refactored the approach based on our reproduction findings.

### Root cause (confirmed with a reproduction script)
PHPWord's HTML writer embeds images as base64 data URIs on a single HTML line — no newlines in the base64 string. When the image is large enough, the line itself exceeds `pcre.backtrack_limit` (1,000,000 chars by default). Mpdf performs a preemptive length check (`strlen($html) > $limit`) before any PCRE runs and throws the error. Because the image is all on one line, neither the original line-by-line loop nor a simple 1,000-line chunk approach is sufficient on its own — both still produce a `WriteHTML` call containing that single oversized line.

### The fix (two parts)
1. Adopt the 1,000-line batching approach (mirrors `PhpSpreadsheet` alignment).
2. When a chunk still exceeds `pcre.backtrack_limit` — because it contains a large base64 data URI — temporarily raise the limit to accommodate that chunk, then restore it immediately. Mpdf's check is a conservative length guard; the actual PCRE patterns do not backtrack deeply on base64 attribute values, so real backtracking failures do not occur.

### Reproduction script (not committed — for reviewer reference)
```php
<?php
use PhpOffice\PhpWord\PhpWord;
use PhpOffice\PhpWord\Settings;

// Random-noise JPEG resists JPEG compression -> ~1.7 MB on disk -> triggers the error
$img = imagecreatetruecolor(1500, 1000);
for ($y = 0; $y < 1000; $y++) {
    for ($x = 0; $x < 1500; $x++) {
        imagesetpixel($img, $x, $y, imagecolorallocate($img, rand(0,255), rand(0,255), rand(0,255)));
    }
}
ob_start(); imagejpeg($img, null, 95); $imageData = ob_get_clean(); imagedestroy($img);
$tmpImage = sys_get_temp_dir() . '/phpword_test.jpg';
file_put_contents($tmpImage, $imageData);
echo 'Image: ' . number_format(strlen($imageData)/1024, 1) . " KB\n";

Settings::setPdfRenderer(Settings::PDF_RENDERER_MPDF, realpath(__DIR__ . '/vendor/mpdf/mpdf'));
$phpWord = new PhpWord();
$phpWord->addSection()->addImage($tmpImage, ['width' => 400, 'height' => 267]);
try {
    IOFactory::createWriter($phpWord, 'PDF')->save(sys_get_temp_dir() . '/out.pdf');
    echo "SUCCESS\n";
} catch (\Throwable $e) {
    echo "FAILED: " . $e->getMessage() . "\n";
}
unlink($tmpImage);
Before the fix: FAILED: The HTML code size is larger than pcre.backtrack_limit 1000000

After the fix: SUCCESS — PDF written (1,740 KB)

The PR also adds a regression unit test (testLargeImageDoesNotExceedBacktrackLimit) that generates a large image and asserts the save succeeds at the PHP default pcre.backtrack_limit without any external ini_set workaround.


idan57570idan-svg:
"Closing this in favor of the clean implementation in #2879."

PR #2890, opened 2026-07-09
Fixed errors by phpstan

pwsdotru:
### Description

Fixed errors from phpstan

Now in draft. Fixed some error in tests. 
To tests added dataProviders insted foreach in test methods.

Used directive @runTestsInSeparateProcesses for back compatible. Because mixed type definition is on PHP >= 8.0

### Checklist:

- [x] My CI is :green_circle:

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -24,8 +24,6 @@ parameters:
         ## <=PHP7.3
         -
             identifier: nullCoalesce.offset
-        -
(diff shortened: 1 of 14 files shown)


pwsdotru:
@Progi1984 can you see this pull request, please.

There are main changes and tests and only one little fix in src.  

And as result we will have green CI ;-)

Progi1984 (approved):

PR #2899, opened 2026-08-15
Fix: <p> wrapping a block element throws "Cannot add ListItemRun in TextRun"

aswinkumar863:
## Summary
`Html::parseParagraph()` always converts a `<p>` into a `TextRun`, but a `TextRun`
cannot contain block-level elements such as `ListItemRun` or `Table`. HTML that
wraps a `<ul>`/`<ol>`/`<table>`/heading in a `<p>` — common output from rich text
editors like TinyMCE — throws `Cannot add ListItemRun in TextRun` and aborts the
whole conversion, rather than just rendering incorrectly.

`parseCell()` already has a `shouldAddTextRun()` guard for this exact constraint;
`parseParagraph()` now reuses it and stays transparent when the paragraph contains
a block element, so its children are added directly to the parent container instead.

## Test plan
- [x] Added `testParseParagraphWrappingList()` covering `<p><ul>...</ul></p>`
- [x] `composer test-no-coverage` — full suite passes
- [x] `composer phpstan` — clean
- [x] `composer phpmd` — clean
- [x] PHP CS Fixer — no violations

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -304,13 +304,20 @@ class Html
      * @param AbstractContainer $element
      * @param array &$styles
      *
-     * @return \PhpOffice\PhpWord\Element\PageBreak|TextRun
+     * @return AbstractContainer|\PhpOffice\PhpWord\Element\PageBreak|TextRun
      */
     protected static function parseParagraph($node, $element, &$styles)
     {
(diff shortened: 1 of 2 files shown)


Progi1984 (commented):
Can you modify the CHANGELOG ? (and it will be ok)

PR #2901, opened 2026-08-20
Add TemplateProcessor::macroExists() to check macro presence (#2895)

### LINKED ISSUE #2895: Add MacroExists public method to TemplateProcessor.php
### Describe the problem

I would like to check if a macro exists in a document in order to make a decision. For example if the macro exists, do this stuff, if it doesn't, do this other stuff.

I don't believe there is a way accomplish this at present so I have written a simple public method called macroExists for TemplateProcessor.php in the next section for this purpose - feel free to rename/modify it as you see fit (e.g. renaming to isMacro).

### Describe the expected behavior

I believe adding the following simple macroExists public method to TemplateProcessor.php will suffice and appears to be working, it returns true if the macro/placeholder text is found, false otherwise:

```
public function macroExists($macro)
{
    return ($this->findMacro($macro) >= 0) ? true : false;
}
```

As you can see it takes advantage of the existing protected findMacro method.

I have tried to create a pull request for this but was unable.

### Priority

- [ ] I want to crowdfund the feature (with [@algora-io](https://docs.algora.io/bounties/overview)) and fund a community developer.
- [ ] I want to pay the feature and fund a maintainer for that. (Contact @Progi1984)

98n:
Adds `TemplateProcessor::macroExists($macro)` so callers can branch on whether a placeholder exists before setting or cloning it. Builds on the existing protected `findMacro()` helper (returns -1 when absent).

Regression test added in `TemplateProcessorTest::testMacroExists`.

Closes #2895.

---
**Algora bounty:** PHPWord issue #2895 (https://github.com/PHPOffice/PHPWord/issues/2895). Submitted for the Algora-funded feature request.

### DIFF
--- a/src/PhpWord/TemplateProcessor.php
+++ b/src/PhpWord/TemplateProcessor.php
@@ -1357,6 +1357,21 @@ class TemplateProcessor
     }
 
     /**
+     * Check whether a macro exists in the document.
+     *
+     * Useful to branch logic on the presence of a placeholder before
+     * setting or cloning it.
+     *
+     * @param string $macro Name of the macro
+     *
+     * @return bool True if the macro is present, false otherwise
+     */
+    public function macroExists($macro)
+    {
+        return $this->findMacro($macro) >= 0;
+    }
+
+    /**
      * Find start and end of XML block containing the given macro
      * e.g. <w:p>...${macro}...</w:p>.
      *
--- a/tests/PhpWordTests/TemplateProcessorTest.php
+++ b/tests/PhpWordTests/TemplateProcessorTest.php
@@ -482,6 +482,18 @@ final class TemplateProcessorTest extends \PHPUnit\Framework\TestCase
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/PhpWord/TemplateProcessor.php
```diff
@@ -1356,6 +1356,21 @@ public function replaceXmlBlock($macro, $block, $blockType = 'w:p')
...
+     * Check whether a macro exists in the document.
+     *
+     * Useful to branch logic on the presence of a placeholder before
+     * setting or cloning it.
+     *
+     * @param string $macro Name of the macro
+     *
+     * @return bool True if the macro is present, false otherwise
+     */
+    public function macroExists($macro)
```
Progi1984, line 1369:
```suggestion
    public function macroExists(string $macro): bool
```

Progi1984 (commented):
@98n Can you fix my feedback and update the changelog ?

PR #2903, opened 2026-08-23
Fix various CSS border parsing and rendering bugs in HTML-to-Word conversion

aswinkumar863:
## Summary
Several related bugs in `Html::parseStyleDeclarations()` and the Word2007 table
border writer meant CSS borders frequently rendered wrong, or not at all:

- The `border`/`border-top`/`border-right`/`border-bottom`/`border-left`
  shorthand assumed a fixed `width color style` token order, but real CSS
  (and every browser/editor) emits `width style color`. Tokens are now
  classified by shape (a number+unit is the width, a recognized keyword is
  the style, anything else is the color) instead of by position, so either
  order parses correctly.
- `border-color` decided single- vs. multi-color by counting `#` characters
  in the value, so named colors (e.g. `red`) had zero and were silently
  dropped entirely, for both the single- and multi-value cases.
- `border-width` was converted with `Converter::cssToPoint()` but stored
  and read back everywhere else in the file as twips, rendering roughly
  20x thinner than intended.
- `border-style` only recognized `none`/`dashed`/`dotted`/`double`, silently
  falling back to `single` (solid) for `groove`/`ridge`/`inset`/`outset`.
- `Style\Border::hasBorder()` - which every writer checks before emitting a
  border at all - gates purely on border size being set. So a bare
  `border-style`/`border-color` declaration without an explicit width was
  dropped entirely, even though CSS itself defaults an unset border-width
  to "medium" (~3px) whenever a style/color is present. The HTML parser now
  backfills that same default so it isn't dropped.
- `Writer\Word2007\Style\Table::writeBorder()` never passed border *style*
  through to its `MarginBorder` XML writer (only size and color), even
  though `MarginBorder` supports it and the cell writer
  (`Writer\Word2007\Style\Cell`) already does. So a table's own
  `<w:tblBorders>` always rendered every side as `single`, regardless of
  what `border-style` was configured - only cell-level borders (`<w:tcBorders>`)
  ever showed the right style.

These are grouped into one PR because they're all in the same small area of
code (the border cases in `parseStyleDeclarations()`, plus `mapBorderStyle()`/
`mapBorderColor()`) and only become fully visible together - e.g. a plain
`border-style: dashed` only round-trips to Word once both the backfill and
the table writer fix are in place.

## Test plan
- [x] Added 6 new tests to `HtmlTest.php` covering: order-independent shorthand
      parsing (hex + named color), named `border-color`, `border-width`
      longhand/shorthand unit consistency, extended `border-style` keywords,
      style-without-width backfill, and table-level `border-style` writing
- [x] `composer test-no-coverage` - full suite passes (1119 tests)
- [x] `composer phpstan` - clean
- [x] `composer phpmd` - clean
- [x] PHP CS Fixer - no violations
- [x] Manually generated a `.docx` exercising every fix and visually confirmed
      correct rendering in Word

### DIFF
--- a/src/PhpWord/Shared/Html.php
+++ b/src/PhpWord/Shared/Html.php
@@ -28,6 +28,7 @@ use PhpOffice\PhpWord\Element\Row;
 use PhpOffice\PhpWord\Element\Table;
 use PhpOffice\PhpWord\Element\TextRun;
 use PhpOffice\PhpWord\Settings;
+use PhpOffice\PhpWord\SimpleType\Border as BorderType;
 use PhpOffice\PhpWord\SimpleType\Jc;
 use PhpOffice\PhpWord\SimpleType\NumberFormat;
 use PhpOffice\PhpWord\Style\Paragraph;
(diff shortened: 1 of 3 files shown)


Progi1984:
@aswinkumar863 Could you rebase?

aswinkumar863:
@Progi1984 Rebased onto the latest master, no conflicts.

Progi1984 (commented):
@aswinkumar863 Could you update the changelog plz ?

PR #2912, opened 2026-08-25
Writer ODText: support native bookmarks

### LINKED ISSUE #2911: Writer ODText: support native bookmarks
## Problem

The ODText writer silently drops PHPWord bookmark elements, even though bookmarks are supported by the PHPWord document model and serialized by the Word2007 writer.

## Expected behavior

ODText should serialize point bookmarks using the native ODF `text:bookmark` element and preserve the bookmark name.

## Scope

- Support bookmarks in ordinary ODText containers, including nested text runs.
- Preserve multiple bookmark names and XML-safe escaping through the existing XML writer.
- Add focused regression tests against generated `content.xml`.

Cross-reference fields and range bookmarks are outside this issue because the current PHPWord bookmark model exposes point bookmarks only.

potofcoffee:
## Description

PHPWord's ODText writer silently drops bookmark elements even though bookmarks are part of the format-neutral model and are supported by the Word2007 writer. This PR serializes point bookmarks as native ODF bookmarks.

Fixes #2911

## Native ODF mapping

- Point bookmarks use the standard `text:bookmark` element with the PHPWord bookmark name.
- The writer preserves bookmarks in ordinary containers and nested text runs.
- Bookmark ranges and cross-reference fields remain outside this change because the current PHPWord bookmark model exposes point bookmarks only.

## Tests

- Added direct `content.xml` regression coverage for single and multiple bookmarks.
- Covered nested text-run placement and XML-safe bookmark names.
- Ran the focused ODText tests, the complete ODText suite, PHP-CS-Fixer, PHPMD, PHPStan, and `git diff --check`.

## Checklist

- [x] My CI is :green_circle:
- [x] I have covered by unit tests my new code
- [x] I have updated the documentation
- [x] I have updated the changelog

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -119,7 +119,7 @@
         "friendsofphp/php-cs-fixer": "^3.3",
         "mpdf/mpdf": "^7.0 || ^8.0",
         "phpmd/phpmd": "^2.13",
-        "phpstan/phpstan": "^0.12.88 || ^1.0.0 || ^2.0.0",
+        "phpstan/phpstan": "^0.12.88 || ^1.0.0 || >=2.0.0 <2.2.11",
         "phpstan/phpstan-phpunit": "^1.0 || ^2.0",
         "phpunit/phpunit": "^7.0 || ^8.0 || ^9.0 || ^10.0 || ^11.0",
         "symfony/process": "^4.4 || ^5.0",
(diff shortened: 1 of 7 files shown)


Progi1984 (approved):

Progi1984:
@potofcoffee Can you rebase ?

Progi1984 (approved):

PR #2930, opened 2026-08-30
Fix HTML strikethrough tags and decimal percentage table widths

aswinkumar863:
## Summary
Two small, unrelated parsing bugs in `Html.php`, bundled together since each is a
one-line fix on its own:

- `<s>`, `<strike>` and `<del>` were not mapped to strikethrough. The parser only
  recognized strikethrough via CSS `text-decoration: line-through`, not any of the
  three semantically-strikethrough tags - so strikethrough text produced by rich
  text editors (e.g. TinyMCE emits `<s>`) silently rendered as plain text.
- The CSS `width` percentage regex `/([0-9]+)%/` had no decimal point, so a value
  like `49.85%` matched only the digits immediately before the `%` (`"85"`),
  silently turning it into 85% instead of ~50%. This affects any table/cell using
  a non-integer percentage width - which is common when widths are computed
  programmatically (e.g. `width.toFixed(2) + '%'`).

## Test plan
- [x] Added `testParseStrikethroughTags` (data provider over `s`/`strike`/`del`)
      and `testParseDecimalPercentageWidth` to `HtmlTest.php`
- [x] `composer test-no-coverage` - full suite passes (1132 tests)
- [x] `composer phpstan` - clean
- [x] `composer phpmd` - clean
- [x] PHP CS Fixer - no violations
- [x] Manually generated a `.docx` exercising both fixes and visually confirmed
      correct rendering in Word

### DIFF
--- a/docs/changes/1.x/1.5.0.md
+++ b/docs/changes/1.x/1.5.0.md
@@ -14,6 +14,7 @@
 
 - Set writeAttribute return type by [@radarhere](https://github.com/radarhere) fixing [#2204](https://github.com/PHPOffice/PHPWord/issues/2204) in [#2776](https://github.com/PHPOffice/PHPWord/pull/2776)
 - Fix "Using null as an array offset" deprecation in Style::getStyle() on PHP 8.5 by [@pwsdotru](https://github.com/pwsdotru) & [@slayerfx](https://github.com/slayerfx) fixing [#2863](https://github.com/PHPOffice/PHPWord/issues/2863) in [#2892](https://github.com/PHPOffice/PHPWord/pull/2892)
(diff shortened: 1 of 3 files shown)


Progi1984 (commented):
@aswinkumar863 Can you update the changelog ?

aswinkumar863:
@Progi1984 Changelog's updated.

This looks like flaky CI, unrelated to the change. Could you re-run it when you get a chance?

Progi1984 (approved):

Progi1984 (approved):

Progi1984:
Thanks @aswinkumar863 for your contribution

